PR #14, opened 2021-03-18
Specify the notebook directory explicitly

### LINKED ISSUE #13: Run commands from outside a notebook directory
[There has been some discussions](https://github.com/mickael-menu/zk/issues/10) around the best way to manage a global notebook.

Declaring in the global configuration file a default notebook would go against the design of `zk` in my opinion. A notebook is like a Git repository, with a `.zk` marker similar to `.git`.

However, it could be interesting to specify from the command line the notebook where `zk` commands are run. There is a precedent with Git, which provides [`--git-dir`](https://git-scm.com/docs/git#Documentation/git.txt---git-dirltpathgt) and [`--work-tree`](https://git-scm.com/docs/git#Documentation/git.txt---work-treeltpathgt).

I suggest specifying the path to the notebook with (in order of precedence):

* `--notebook-dir <path>`
* current working directory (or its parents)
* `ZK_NOTEBOOK_DIR` env variable

To simplify, any relative path provided as arguments would be relative to the notebook directory (when set explicitly), which allows stuff like:

```sh
$ zk list --exclude journal --notebook-dir ~/notes
```

In practice, that means that someone using a single notebook can set it as a "default global one" with `ZK_NOTEBOOK_DIR` in `~/.profile`.

mickael-menu:
Fix #13

* Use `--notebook-dir` or set `ZK_NOTEBOOK_DIR` to run `zk` as if it was started from this path instead of the current working directory.
    * This allows running `zk` without being in a notebook.
    * By setting `ZK_NOTEBOOK_DIR` in your shell configuration file (e.g. `~/.profile`), you are declaring a default global notebook which will be used when `zk` is not in a notebook.
    * When the notebook directory is set explicitly, any path given as argument will be relative to it instead of the actual working directory.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ All notable changes to this project will be documented in this file.
 * Use `--notebook-dir` or set `ZK_NOTEBOOK_DIR` to run `zk` as if it was started from this path instead of the current working directory.
     * This allows running `zk` without being in a notebook.
     * By setting `ZK_NOTEBOOK_DIR` in your shell configuration file (e.g. `~/.profile`), you are declaring a default global notebook which will be used when `zk` is not in a notebook.
+    * When the notebook directory is set explicitly, any path given as argument will be relative to it instead of the actual working directory.
 
 
 ## 0.2.1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ All notable changes to this project will be documented in this file.
     * This is the same format as a notebook [configuration file](docs/config.md).
     * Shared templates can be stored in `~/.config/zk/templates/`.
     * `XDG_CONFIG_HOME` is taken into account.
+* Use `--notebook-dir` or set `ZK_NOTEBOOK_DIR` to run `zk` as if it was started from this path instead of the current working directory.
+    * This allows running `zk` without being in a notebook.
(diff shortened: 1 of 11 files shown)


megalithic:
@mickael-menu just to confirm.. let me know if i'm missing something else..

image

mickael-menu:
@megalithic How did you build `zk`, from a clone? You can check if you are up to date if you see this option with `zk -h`:

```
      --notebook-dir=PATH    Run as if zk was started in <PATH> instead of the current working directory.
```

[Here's a build I made for macOS](https://github.com/mickael-menu/zk/files/6166701/zk.zip)

mickael-menu:
I just pushed a fix on `main`, seems like an import was lost from the PR merge.

Also if you build with the `./go` script in the repo, you get a more precise version number.

Screenshot 2021-03-18 at 20 27 52

megalithic:
> Ha I guess you used `go get -u` from your plugin. The problem is that this will fetch the last tagged version, but I didn't release yet these changes. You can add `@HEAD` to get the latest commit instead.
> 
> Also you will need to pass the `fts5` and `icu` tags, otherwise many SQLite features won't work in your build.
> 
> ```shell
> go get -tags "fts5 icu"  -u github.com/mickael-menu/zk@HEAD
> ```
> 
> The `./go` script contains these tags, and if you use it you also get a nicer version number:
> 
> Screenshot 2021-03-18 at 20 27 52

You nailed it! Thank you! Will try again when I get back from a meeting.

megalithic:
this works! so good! thanks @mickael-menu! hoping to add more of our discussed features/functionality soon.

PR #15, opened 2021-03-20
Add the --mentioned-by filtering option

mickael-menu:
* Find every note whose title is mentioned in the note you are working on with `--mentioned-by file.md`.
    * To refer to a note using several names, you can use the [YAML frontmatter key `aliases`](https://publish.obsidian.md/help/How+to/Add+aliases+to+note). For example the note titled "Artificial Intelligence" might have: `aliases: [AI, robot]`
    * To find only unlinked mentions, pair it with `--no-linked-by`, e.g. `--mentioned-by file.md --no-linked-by file.md`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ All notable changes to this project will be documented in this file.
     * This allows running `zk` without being in a notebook.
     * By setting `ZK_NOTEBOOK_DIR` in your shell configuration file (e.g. `~/.profile`), you are declaring a default global notebook which will be used when `zk` is not in a notebook.
     * When the notebook directory is set explicitly, any path given as argument will be relative to it instead of the actual working directory.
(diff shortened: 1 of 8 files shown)


mickael-menu:
@megalithic @evantravers I'm pretty excited about this one, which is the inverse of the `--mention` option. I think it might be more useful when writing a note, to find existing notes to link to. As far as I know Obsidian doesn't have this feature yet.

mickael-menu:
I'm wondering if we could have a way to highlight unlinked mentions in real time with nvim, then automatically create a link with some  key mapping when over such mention. A bit like in TheBrain 12 in this video excerpt: https://youtu.be/9eeuUWrYQPE?t=57

I will think about a more parser-friendly output for `zk list`. Maybe JSON could be relevant since it is supported out of the box by nvim.

evantravers:
That is very cool.

PR #38, opened 2021-05-05
lsp: parse URIs properly to allow spaces in paths

cormacrelf:
For paths like `/path to my/zettelkasten`. Uses the `url` package from Go.

### DIFF
--- a/internal/adapter/lsp/server.go
+++ b/internal/adapter/lsp/server.go
@@ -143,7 +143,12 @@ func NewServer(opts ServerOpts) *Server {
 			return nil
 		}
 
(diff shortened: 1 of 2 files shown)


mickael-menu (approved):
Thanks, using the `url` package is indeed much better!

Would you mind checking the "allow edits from maintainers" if you open another PR in the future? I'm updating the changelog in the original PR usually. This gives me write access only to the PR branch, not your whole fork.

cormacrelf:
It seems to be checked by default? I didn't do it deliberately for either but my other PR has it enabled!

mickael-menu:
Yes I think it is enabled by default, weird.

PR #39, opened 2021-05-05
lsp: don't null-pointer-deref panic when TextDocumentCompletion.Context missing

cormacrelf:


internal/adapter/lsp/server.go
```diff
@@ -174,9 +174,13 @@ func NewServer(opts ServerOpts) *Server {
 	}
 
 	handler.TextDocumentCompletion = func(context *glsp.Context, params *protocol.CompletionParams) (interface{}, error) {
-		triggerChar := params.Context.TriggerCharacter
-		if params.Context.TriggerKind != protocol.CompletionTriggerKindTriggerCharacter || triggerChar == nil {
-			return nil, nil
+		triggerChar := "["
```
mickael-menu, line 177: The auto-completion is also used for tags, so I'm afraid that defaulting to `[` would make it unreachable. Maybe we could just take a look at the previously entered characters ([like here](https://github.com/mickael-menu/zk/blob/9df2cb25ac6053cbd81f04c2839c22a0fd08e99a/internal/adapter/lsp/server.go#L206)) and ignore the trigger character in this case?

I didn't encounter an editor which doesn't support the context and trigger characters yet. How is the auto-completion triggered in this case, manually with C-Space? Which editor / LSP plugin are you using?

cormacrelf: Neovim's built in LSP doesn't do it yet. Moreover the spec says it's optional, so you do have to check the nil. I think it is perhaps a low priority feature, because generally LSPs are accepting file content as well as completion requests, and have access to a stored copy of the file. So nobody actually needs to know the trigger character. Not even you! I'll fix with a look behind instead.

cormacrelf: It's also optional because completion gets triggered in many other ways than typing a character.

mickael-menu: Yeah in the end this context is not super useful then. Thanks for addressing this!

mickael-menu:
Superseded by https://github.com/mickael-menu/zk/pull/41

@cormacrelf I took over this PR in another branch (as I didn't have write access on yours) but credited you in the changelog. Thanks again for spotting this 🙏

mickael-menu:
Superseded by https://github.com/mickael-menu/zk/pull/41

@cormacrelf I took over this PR in another branch (as I didn't have write access on yours) but credited you in the changelog. Thanks again for spotting this 🙏

PR #57, opened 2021-06-29
remove go chmod command from README.md

Alys:
Now that the `go` script no longer exists in the repo, the chmod command isn't needed.

BTW zk is a great tool, I'm absolutely loving it. Thanks!


mickael-menu (approved):
Thanks for spotting that, and the kind words!

PR #58, opened 2021-07-03
lsp: add references support for links

pstuifzand:
The LSP returns a list of references to the link that is pointed to.

### DIFF
--- a/internal/adapter/lsp/server.go
+++ b/internal/adapter/lsp/server.go
@@ -439,7 +439,7 @@ func NewServer(opts ServerOpts) *Server {
 			}
 
 			locations = append(locations, protocol.Location{
-				URI: "file://" + filepath.Join(notebook.Path, note.Path),
+				URI: pathToURI(filepath.Join(notebook.Path, note.Path)),
 				Range: protocol.Range{
 					Start: protocol.Position{
 						Line:      line,
--- a/internal/adapter/lsp/server.go
+++ b/internal/adapter/lsp/server.go
@@ -104,11 +104,7 @@ func NewServer(opts ServerOpts) *Server {
 			ResolveProvider:   boolPtr(true),
 		}
 
-		capabilities.ReferencesProvider = &protocol.ReferenceOptions{
-			WorkDoneProgressOptions: protocol.WorkDoneProgressOptions{
-				WorkDoneProgress: boolPtr(false),
-			},
-		}
+		capabilities.ReferencesProvider = &protocol.ReferenceOptions{}
 
 		return protocol.InitializeResult{
 			Capabilities: capabilities,
@@ -416,17 +412,13 @@ func NewServer(opts ServerOpts) *Server {
 			return nil, err
 		}
 
-		opts := core.NoteFindOpts{}
 		p, err := notebook.RelPath(target.Path)
 		if err != nil {
 			return nil, err
 		}
 
-		opts.LinkTo = &core.LinkFilter{
-			Paths:       []string{p},
-			Negate:      false,
-			Recursive:   true,
-			MaxDistance: 1,
+		opts := core.NoteFindOpts{
+			LinkTo: &core.LinkFilter{Paths: []string{p}},
 		}
 
 		notes, err := notebook.FindNotes(opts)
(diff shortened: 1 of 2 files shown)


mickael-menu:
@pstuifzand Thanks, that's a useful feature!

I'd like to update the changelog and simplify a bit the code but I'm not allowed to push on your branch. Would you mind enabling write access from me? Thanks.

https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

internal/adapter/lsp/server.go
```diff
@@ -104,6 +104,12 @@ func NewServer(opts ServerOpts) *Server {
 			ResolveProvider:   boolPtr(true),
 		}
 
+		capabilities.ReferencesProvider = &protocol.ReferenceOptions{
+			WorkDoneProgressOptions: protocol.WorkDoneProgressOptions{
+				WorkDoneProgress: boolPtr(false),
+			},
+		}
```
mickael-menu, line 111: These are the default values so we can skip it.
```suggestion
		capabilities.ReferencesProvider = &protocol.ReferenceOptions{}
```

internal/adapter/lsp/server.go
```diff
@@ -389,6 +395,75 @@ func NewServer(opts ServerOpts) *Server {
...
+		if err != nil {
+			return nil, err
+		}
+
+		opts.LinkTo = &core.LinkFilter{
+			Paths:       []string{p},
+			Negate:      false,
+			Recursive:   true,
+			MaxDistance: 1,
+		}
```
mickael-menu, line 430: `Recursive` can be used to look for indirect links across several notes. But since you used a `MaxDistance` of 1, it's like having recursive disabled, which is the default. It's a bit more efficient with SQLite to disable recursive when not needed.

```suggestion
		p, err := notebook.RelPath(target.Path)
		if err != nil {
			return nil, err
		}

		opts := core.NoteFindOpts{
			LinkTo: &core.LinkFilter{Paths: []string{p}},
		}
```

internal/adapter/lsp/server.go
```diff
@@ -389,6 +395,75 @@ func NewServer(opts ServerOpts) *Server {
...
+			var line uint32 = 0
+			if pos < 0 {
+				line = 0
+			} else {
+				linePos := strings.Count(note.RawContent[0:pos], "\n")
+				line = uint32(linePos)
+			}
+
+			locations = append(locations, protocol.Location{
+				URI: "file://" + filepath.Join(notebook.Path, note.Path),
```
mickael-menu, line 450: Prefer `pathToURI` to generate the file URIs because it will handle encoding special characters and spaces.

```suggestion
				URI: pathToURI(filepath.Join(notebook.Path, note.Path)),
```

internal/adapter/lsp/server.go
```diff
@@ -389,6 +395,75 @@ func NewServer(opts ServerOpts) *Server {
...
+
+		notes, err := notebook.FindNotes(opts)
+		if err != nil {
+			return nil, err
+		}
+
+		var locations []protocol.Location
+
+		for _, note := range notes {
+			pos := strings.Index(note.RawContent, target.Path[0:len(target.Path)-3])
```
mickael-menu, line 440: I expect this will fail if the notes are not in the same root folder, as the `target.Path` is always relative to the root of the notebook. The `-3` might be an issue too when not using file extensions.

However, worst case scenario is that we will preview the first line of the note, so I think it's fine for now. I'm planning to add an API to get the actual link infos which will provide the exact link position, we can refactor this then.

mickael-menu (changes requested):

internal/adapter/lsp/server.go
pstuifzand: It would be useful to have the actual position of the links in the files.

pstuifzand:
I have made the changes you requested.

internal/adapter/lsp/server.go
mickael-menu: Yep definitely. Right now in the database we have the position of the snippet containing the link, but I think I will add the link position itself as well.

mickael-menu (approved):
Perfect, thanks.

PR #60, opened 2021-07-10
Add "abs-path" to template-format

pstuifzand:
Adds a full path for the note to the list template to make integration easier.

### DIFF
--- a/docs/template-format.md
+++ b/docs/template-format.md
@@ -5,7 +5,7 @@ The following variables are available in the templates used when formatting note
 | Variable      | Type     | Description                                                              |
 |---------------|----------|--------------------------------------------------------------------------|
 | `path`        | string   | File path to the note, relative to the current directory                 |
-| `full-path`   | string   | File path to the note, full path including the notebook directory        |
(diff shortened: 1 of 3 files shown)


docs/template-format.md
```diff
@@ -5,6 +5,7 @@ The following variables are available in the templates used when formatting note
 | Variable      | Type     | Description                                                              |
 |---------------|----------|--------------------------------------------------------------------------|
 | `path`        | string   | File path to the note, relative to the current directory                 |
+| `full-path`   | string   | File path to the note, full path including the notebook directory        |
```
mickael-menu, line 8: Sounds useful 👍 

Could you rename it to `abs-path` instead? That's what I've been using in a similar context: https://github.com/mickael-menu/zk/blob/4d893b9bd6934170d181b5dafe460802700bec1c/docs/tool-fzf.md#template-context

mickael-menu (changes requested):
@pstuifzand Could you check out the unit tests? Seems like they don't pass anymore. Thanks.

```
make test
```

mickael-menu (approved):

mickael-menu:
Thanks!

PR #104, opened 2021-11-13
Fix indexing wiki links using partial paths

### LINKED ISSUE #98: zk does not recognize my Wiki link template with a pipe
Here is part of my configuration:

```toml
[note]
filename = "{{slug title}}.{{id}}"

[format.markdown]
link-format = "[[{{metadata.id}}|{{title}}]]"
```

This generates a text link from the LSP something like this:

```
[[ker6z|Terms of Gardening]]
```

But it does not get recognized as an actual link to the note:

```bash
$ zk list -l Digital\ Garden/terms-of-gardening.ker6z.md
Found 0 note
```

My intention with the template is to be able to link to a template without worrying what its filename is as well as enabling custom link text in the source document to allow better flow in whatever context it is in.

For me, when "wiki" is listed as a supported link style, I expected nearly-full support of [MediaWiki's link syntax](https://en.wikipedia.org/wiki/Help:Link) but it seems it's much simpler then that. Could this feature be added?

mickael-menu:
Fix #98 

### Fixed

* [#98](https://github.com/mickael-menu/zk/issues/98) Index wiki links using partial paths for `--linked-by` and `--link-to`.
* [#98](https://github.com/mickael-menu/zk/issues/98) Ignore spaces around the pipe in wiki links for LSP diagnostics.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ All notable changes to this project will be documented in this file.
 ### Fixed
 
 * [#89](https://github.com/mickael-menu/zk/issues/89) Calling `zk index` from outside the notebook (contributed by [@adamreese](https://github.com/mickael-menu/zk/pull/90)).
+* [#98](https://github.com/mickael-menu/zk/issues/98) Index wiki links using partial paths for `--linked-by` and `--link-to`.
 
 
 ## 0.7.0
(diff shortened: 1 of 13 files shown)


nogweii:
nice, this works very well. Thank you for taking a look at it!

PR #142, opened 2022-01-05
Update installation instruction for MacOS

tranvnhan:
The square brackets are not necessary and may cause problems for anyone who is not familiar with the Brew package manager.


mickael-menu:
I added the square brackets to show that the flag is optional. Granted this is maybe not very obvious, I'll update the README.

Issue #137, opened 2021-12-31
Allow different options for FzF in interactive filtering


Nelyah:
Thank you for  the tool, it is a brilliant piece of software which I'm enjoying very much!

One of thing I wish would be possible is adjusting the FZF parameters. The one I'm thinking most is `--height 40%` (instead of 100% in `zk edit -i`). I've looked a bit at the code and it seems to be set in `internal/adapter/fzf/fzf.go:75`. 

One solution would be to prepend that list of arguments with the env variable `$FZF_DEFAULT_OPTS` if it exists, and append to the list custom options that could be defined in the `.zk/config.toml`. This would make the default config for fzf in zk override the default `fzf` config (makes sense to me) but would still allow users to override the `zk` default config (`fzf --height 100% --height 40%` results in height at 40%).

Do you think this would be possible?

mickael-menu:
> Thank you for the tool, it is a brilliant piece of software which I'm enjoying very much!

Thank you for the kind words!

> One solution would be to prepend that list of arguments with the env variable `$FZF_DEFAULT_OPTS` if it exists, and append to the list custom options that could be defined in the `.zk/config.toml`. This would make the default config for fzf in zk override the default `fzf` config (makes sense to me) but would still allow users to override the `zk` default config (`fzf --height 100% --height 40%` results in height at 40%).

That sounds like a good strategy. The user options could be specified [in the configuration file](https://github.com/mickael-menu/zk/blob/main/docs/tool-fzf.md#integration-with-fzf):

```toml
[tool]
fzf-options = "--height 40%"
```

Would you be interested in contributing a PR for this?

Nelyah:
I will try to see in the next few days if I can come up with something, though I have never really written a single line of Go before. 

The issue doesn't seem _too_ complicated though, so maybe I'll be able to get something working and this would be a good introduction to the language. I'll let you know how this goes!

mickael-menu:
Thanks, you can create a draft PR and if you check the "Allow edit from maintainers" checkbox while creating the PR I can finish it for you if needed.

mickael-menu:
[`fzf-options`](https://github.com/mickael-menu/zk/blob/main/docs/tool-fzf.md) is now available in `main`.

PR #154, opened 2022-01-14
Make FzF options configurable

Nelyah:
This adds the ability to configure the FzF options used by the
NewNoteFilter.

Here are the sources for the FzF options, in order of precedence (first
precedes the second):
1. tool.fzf-additional-args in the note directory config file
2. Default arguments set by zk in `internal/adapter/fzf/fzf.go`
3. Environment variable FZF_DEFAULT_OPTS

This should solve #137 

Let me know if I can improve on it. As I mentioned in the issue those are my first lines of Go so it's still a bit of a struggle.


internal/adapter/fzf/fzf.go
```diff
@@ -72,6 +74,8 @@ func New(opts Opts) (*Fzf, error) {
 		opts.Delimiter = "\x01"
 	}
 
+    defaultFzfArgs := strings.Split(os.Getenv("FZF_DEFAULT_OPTS"), " ")
```
mickael-menu, line 77: Here and also line 96, you can use `shellquote.Split()` instead, this will make sure that if one term is between quotes it won't be split at the inner space characters.

It returns an error, in Go you need to check manually if the error is null, and return if needed:

```suggestion
    rawDefaultOpts := os.Getenv("FZF_DEFAULT_OPTS")
    defaultFzfArgs, err := shellquote.Split(rawDefaultOpts)
	if err != nil {
		return nil, errors.Wrapf(err, "can't split the fzf options: %s", rawDefaultOpts)
	}
```

mickael-menu (changes requested):
Thank you for tackling this, I appreciate it!

A few comments:

* I found an issue while testing: if any of `fzf-additional-args` or `FZF_DEFAULT_OPTS` are not set or empty, it will error out with:
    
    ```sh
    $ zk list -i
    unknown option:
    zk: error: failed to filter interactively the output with fzf, try again without --interactive or make sure you have a working fzf installation: exit status 2
    ```
    
    I think using `shellquote` like described in my other comment should solve the problem.
* I don't think you need to add `FZF_DEFAULT_OPTS` manually, it should be handled automatically by `fzf`. I tried `FZF_DEFAULT_OPTS="--border" zk list -i` from the `main` branch and it works.
* I would prefer using `fzf-options` / `FzfOptions` instead of `fzf-additional-args`, to make it closer to `FZF_DEFAULT_OPTS`.
* Some of the `zk` hard-coded options could actually be part of the default value for `fzf-options`, when unset. Some can't be unset by additional arguments (like `--no-hscroll`) and most of them are tuned to my taste. I suggest:
    * Keeping only the essential hard-coded options: `--delimiter` and `--ansi`.
    * Adding to the default value of `fzf-options`: `--tiebreak begin --exact --tabstop 4 --height 100% --layout reverse --no-hscroll --color hl:-1,hl+:-1 --preview-window wrap`
        * Look for `LineTemplate.OrString(defaultLineTemplate).String()` for an example of handling a default value for an option.

Thanks again!

Nelyah:
Thank you very much for the detailed @mickael-menu! I'm sorry I haven't had much time these past few weeks to really get to this, but I haven't forgotten about it :)

megalithic:
Hello friends; just checking in to see if this still has some traction/working towards it. Hope you are all well during these crazy times.

Are there work-arounds for the fzf key binding overrides in the meantime?

mickael-menu:
Hi @megalithic, I intend to finish the work eventually but I've been really busy the past few weeks so I'm not sure when I'll find the time yet.

Nelyah:
Hi, apologies for letting this without update for so long. If you do have the time now to get to it, feel free to. Otherwise I will also try to finish it when I get the time.

megalithic:
whew, glad it's not officially `stale` anymore. 🙏

mickael-menu (approved):
Thank you @Nelyah for kickstarting this PR!

@megalithic I changed the default key binding to create a new note to Ctrl-E to avoid conflicts. But you can choose a different one with:

```toml
[tool]
fzf-bind-new = "Ctrl-C"
```

I plan to make a new release very soon.

PR #160, opened 2022-01-23
Disable `additionalTextEdits` for completion items by default

mickael-menu:
### Added

* [zk-nvim#26](https://github.com/mickael-menu/zk-nvim/issues/26) The LSP server doesn't use `additionalTextEdits` anymore to remove the trigger characters when completing links.
    * You can customize the default behavior with the [`use-additional-text-edits` configuration key](docs/config-lsp.md).


kabouzeid:
For `nvim-cmp` + `delimitMate` setting the new option to either true or false both works.

PR #171, opened 2022-02-22
Fix windows build.

### LINKED ISSUE #139: Feature request: Support for Windows
First off thank you for this great application!

It will be great if Zk supported windows.
I tried to build it on windows but got the following error
```
# github.com/mickael-menu/zk
.\main.go:174:36: invalid use of ... in call to "github.com/mickael-menu/zk/internal/util/exec".CommandFromString
make: *** [Makefile:3: build] Error 2
```

codito:
Use variadic arguments in windows implementation of `CommandFromString`.
Check for windows root directory while searching for notebooks, this
fixes an infinite loop during startup notebook discovery. Fixes #139.

Please note that `go build -tags "fts5 icu"` will require a patch from
`go-sqlite3` [here][] with a msys2/mingw64 env.

[here]: https://github.com/mattn/go-sqlite3/pull/1017

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,5 +14,8 @@
 # Dependency directories (remove the comment below to include it)
 # vendor/
 
+# IDEs/Editors
+.vscode/
+
 notebook.db
 zk
--- a/internal/core/notebook_store.go
+++ b/internal/core/notebook_store.go
@@ -152,7 +152,8 @@ func (ns *NotebookStore) locateNotebook(path string) (string, error) {
 
 	var locate func(string) (string, error)
 	locate = func(currentPath string) (string, error) {
-		if currentPath == "/" || currentPath == "." {
+                // For Windows, the root dir may end with volume name, e.g. E:\\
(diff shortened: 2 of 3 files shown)


dstanberry:
@codito can you clarify if it is possible to use `zk.exe` in powershell/cmd or if it is only expected to work in the mingw64 env?

I've tried using the go-for-windows binary and installing mingw64-go but `zk.exe` only in mingw shell.
It would be nice to be able to interact with the binary in a powershell session and this is probably a noob question but is that the expected outcome?

codito:
@dstanberry it is possible to use zk.exe from powershell. I had to add the mingw64 bin to user PATH env variable. This was needed for go sqlite to find icu dependencies.

dstanberry:
That worked for me, thanks! 

If this hasn't already been discussed, it could be helpful to add some documentation (even if it is unsupported) for other windows users.

Sorry for polluting this chain.

mickael-menu:
> If this hasn't already been discussed, it could be helpful to add some documentation (even if it is unsupported) for other windows users.

Sure, if it's clearly stated that the Windows support is provided on a community basis. Feel free to open a PR with some documentation for Windows.

PR #175, opened 2022-02-26
fix: strip first char for paths with `file:///` in Windows

codito:
Allows autocomplete for paths via lsp.


mickael-menu (approved):
Looking good, thank you.

PR #183, opened 2022-03-22
Add `--id` flag to `zk new`

skbolton:
This allows overriding the id generation step and just use a provided
value instead.

### DIFF
--- a/internal/core/notebook.go
+++ b/internal/core/notebook.go
@@ -109,6 +109,8 @@ type NewNoteOpts struct {
 	Date time.Time
 	// Don't save the generated note on the file system.
 	DryRun bool
+	// Use a provided id over generating one
+	ID string
 }
 
 // ErrNoteExists is an error returned when a note already exists with the
@@ -148,6 +150,15 @@ func (n *Notebook) NewNote(opts NewNoteOpts) (*Note, error) {
 		return nil, wrap(err)
 	}
 
+	var idGenerator IDGenerator
+	if opts.ID != "" {
+		idGenerator = func() string {
+			return opts.ID
+		}
+	} else {
+		idGenerator = n.idGeneratorFactory(config.Note.IDOptions)
+	}
+
 	task := newNoteTask{
 		dir:              dir,
 		title:            opts.Title.OrString(config.Note.DefaultTitle).Unwrap(),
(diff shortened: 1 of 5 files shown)


skbolton:
@mickael-menu new to this project and never done any go before. Would love some feedback on the direction I have gone. I am also wondering if there are other tests that I should write and where they are. Thanks!

tests/fixtures/new/.zk/config.toml
```diff
@@ -10,6 +10,10 @@ filename = "{{date now '%d-%m'}}"
 template = "empty.md"
 filename = "{{now}}"
 
+[group.id.note]
+template = "empty.md"
```
skbolton, line 14: Not partial on how to test the id showing up in the templates. I did a new group to not affect the other tests but maybe just injecting the id into the default template would be better?

internal/core/notebook.go
```diff
@@ -148,6 +150,15 @@ func (n *Notebook) NewNote(opts NewNoteOpts) (*Note, error) {
 		return nil, wrap(err)
 	}
 
+	var idGenerator IDGenerator
+	if opts.ID != "" {
```
skbolton, line 154: The command parsing was confusing me. I was assuming that null values are converted to their type as an empty value. I might be way off though and need to change this.

tests/fixtures/new/.zk/config.toml
mickael-menu: No I think you did well adding a dedicated group.

internal/core/notebook.go
mickael-menu: That's right, there's no `null` values except for pointer types. As you (rightly) used `ID string` instead of `ID *string`, checking for `""` is the correct approach.

You could have used my `opt.String` type as an alternative, but to be honest this is not very idiomatic in Go. I added it as frustration that a real optional Go type was missing. This might change in the new Go 1.18 which added generics.

mickael-menu (approved):
Thank you for tackling this @skbolton!

> @mickael-menu new to this project and never done any go before. Would love some feedback on the direction I have gone.

It's how I would have implemented it, though I'm not a Go expert.

> I am also wondering if there are other tests that I should write and where they are. Thanks!

Not for this feature. I appreciate that you spent the time to add a tesh test case!

PR #186, opened 2022-03-30
Fix file url parsing on Windows

### LINKED ISSUE #184: LSP Hover breaks on windows
I'm using neovim on windows and when I attempt to hover over a tag, the request fails with something similar to the following:
```
zk: -32600: parse "file:///C:%5Cfoo%5Cbar.md": invalid port "%5Cfoo%5Cbar.md"
```

The `%5C` is html escaped 'somewhere' which causes uriToPath(...) to fail:

https://github.com/mickael-menu/zk/blob/a0ced3c33098a813dcb803e350b2ea38aa79ce4b/internal/adapter/lsp/util.go#L19-L23

I've resolved this by doing a strings.ReplaceAll() and can submit a PR if you think this change is acceptable. There's probably a better/cleaner way to do this but I stopped at "this works for me".
```go
package lsp

import(
    ...
    "strings"
    ...
)

...

func uriToPath(uri string) (string, error) {
	s := strings.ReplaceAll(uri, "%5C", "/")
	parsed, err := url.Parse(s)
	if err != nil {
		return "", err
	}
```

dstanberry:
Fixes #184. The changes should only affect Windows builds and will try to resolve file paths when the uri scheme is formatted as either `file:///...` or `file://...`.

### DIFF
--- a/internal/adapter/lsp/util.go
+++ b/internal/adapter/lsp/util.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net/url"
 	"runtime"
+	"strings"
 
 	"github.com/mickael-menu/zk/internal/util/errors"
 )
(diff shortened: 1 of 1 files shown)


mickael-menu (approved):
Thank you 👍

PR #206, opened 2022-05-08
Add Dependabot configuration

artoj:
Dependabot is a service by GitHub that automatically scans the
dependencies of the project. If an outdated dependency is found,
Dependabot will create a PR for updating that dependency.

Add Dependabot configuration that will scan the GitHub workflows and Go
dependencies for possible updates.

Dependabot may have to be manually enabled on the Security tab of the
project.

More information: https://docs.github.com/en/code-security/dependabot


mickael-menu:
Thanks for submitting this PR, however I'm not a big fan of Dependabot and usually prefer updating the dependencies manually.

I would be open to enable it if there's a way to disable the automatic PRs and just keep the warnings.

artoj:
I think `open-pull-requests-limit` is what you want: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#open-pull-requests-limit

Set that to 0 and no PRs will be generated for non-security updates.

Should I update the PR with that value?

mickael-menu (approved):
Thanks @artoj!

PR #213, opened 2022-05-14
Remove dependency on `libicu`

mickael-menu:
### Changed

* Removed the dependency on `libicu`.

---

There were recurring build issues caused by the dependency on `libicu`. Turns out it was not that necessary for `zk`.

* https://github.com/mickael-menu/zk/issues/211
* https://github.com/mickael-menu/zk/issues/189
* https://github.com/mickael-menu/zk/issues/54
* https://github.com/mickael-menu/zk/issues/178
* https://github.com/mickael-menu/zk/issues/53
* https://github.com/mickael-menu/zk/issues/46
* https://github.com/mickael-menu/zk/issues/19


Alys:
Thank you for making this change! I'm now able to use the pre-built binary, whereas before I had to build from source due to a libicuuc.so version issue, so this change is very welcome! <3 :)

mickael-menu:
@Alys You're welcome, this dependency was a real headache! I'm also looking into removing the need for the C compiler altogether which would allow generating fully static binaries.

PR #240, opened 2022-07-10
Support standard input via shell redirection with `zk new`

mickael-menu:
### Fixed

* [#239](https://github.com/mickael-menu/zk/discussions/239) Support standard input via shell redirection with `zk new`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ All notable changes to this project will be documented in this file.
 ### Fixed
 
 * [#233](https://github.com/mickael-menu/zk/issues/233) Hide index progress in non-interactive shells.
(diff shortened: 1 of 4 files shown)


pkazmier:
I submitted #242 with this suggestion. I updated the tests, docs, and made small code change as you'll see. I also renamed `--no-input` to `--no-prompt` as the flag I added was `-I` / `--input`, but I did add make `--no-prompt` backwards compatible by adding an alias for `--no-input` I believe. I know this is your baby, so I won't be offended if you don't care for any of the suggestions. I defer to you.

mickael-menu:
Ha crap, the tesh files didn't pick up on this. I'll answer directly on your PR.

PR #242, opened 2022-07-11
Add explicit flag to read from standard input

pkazmier:
Fixes bug introduced in #240 where interactive use of `zk new` results
in the command sitting indefinitely for user to supply user input.

This change introduces an explicit flag to `new` to indicate standard
input should be read for the content of the note.  The flag is called
`-i` on `--input`.

The flag name is somewhat confusing as there is a `--no-input` flag at
the global level.  I propose that we change that flag name to
`--no-prompt` as it more accurately describes the option.  To maintain
backwards compatibility, we define `--no-input` as an alias.


pkazmier:
Apparently, kong `aliases` are only for commands. Providing a simple 
backwards compatible `--no-input` option  did not work as I had
expected. At this point, `--no-input` has been removed. If we want
to keep it for backwards compatibility, I suppose we could add a new
hidden command line arg called `--no-input`. I defer to you for your
thoughts.

mickael-menu:
Thanks for tackling this!

I'd like to keep the `--no-input` name because it's a pretty standard flag for this. However, if you rename `-i` to `--interactive` it removes the issue while also being consistent with the `--interactive` flag of `zk list`.

This interactive mode for `zk new` is an interesting (albeit unintended!) new feature. But I think we also need to revert to the [`ReadStdinPipe()`](https://github.com/mickael-menu/zk/pull/240/files#diff-988a6e852763f680eba9ae5f9d08cb992f7bc276baf84fb1af725b99b90d25e2L36) mode when `--interactive` is not set.

I'll need to check if there's a way to read all available stdin input without blocking until Ctrl-D, that would be a nicer behavior when `--interactive` is not set.

pkazmier:
> I'd like to keep the `--no-input` name because it's a pretty standard flag for this. However, if you rename `-i` to `--interactive` it removes the issue while also being consistent with the `--interactive` flag of `zk list`.

Yes, `interactive` makes more sense. Not sure what I was thinking tbh.

> This interactive mode for `zk new` is an interesting (albeit unintended!) new feature. But I think we also need to revert to the [`ReadStdinPipe()`](https://github.com/mickael-menu/zk/pull/240/files#diff-988a6e852763f680eba9ae5f9d08cb992f7bc276baf84fb1af725b99b90d25e2L36) mode when `--interactive` is not set.

I don't think it's common for a CLI tool to selectively choose when to read from stdin based on whether the data is coming from a pipe or not, so I'm not sure we need to, but in the spirit of backwards compatibility, it certainly doesn't hurt.
 
> I'll need to check if there's a way to read all available stdin input without blocking until Ctrl-D, that would be a nicer behavior when `--interactive` is not set.

I don't believe there is any standard means to do so.

> By the way, did you change anything in the SVG screencast? It appears as changed in the "Files changed" tab: docs/assets/media/list-format.svg

Yes, in the screen shot, there was a `--no-input` that I changed to `--no-prompt`.  I'll revert all those and make the changes you suggest above tonight or tomorrow morning if you want.

mickael-menu:
Thank you, that would be great 🙏

pkazmier:
Re: tesh, I just realized it was your tool. I should have look at the GitHub workflows earlier. Doh!

pkazmier:
Oh, and to further sweeten the pot, I also updated the screencasts where you pipe content into `zk new` to use the `zk new --interactive` instead. 😄

mickael-menu (approved):

pkazmier:
> > Upon thinking about our earlier discussion, I really believe that `zk new` should only read from standard input, for either shell redirection or shell piping, if told to do so explicitly via the `--interactive` flag. In my opinion, a unix tool should not implicitly read from standard input sometimes, but other times explicitly—it's inconsistent and not the unix way.
> 
> You convinced me, that's a sound argument.

Fantastic!

Issue #243, opened 2022-07-12
LSP Go-to-reference does not work, if notes are in a folder.


bibor:
# Description
The find-references feature of the LSP (tracked in #22) does not work if the notes are organized in folders. 
# Expected Behaviour
find-reference should return all references to the given note.
# Steps to Reproduce

1. have a zettelkasten with the following layout
~~~
$ tree 
.
└── old
    ├── eso7.md
    ├── jbxt.md
    └── vlix.md
~~~
2. set a link from vlix.md to jbxt.md
~~~
# top-note

blub

I reference another note: [ref-note](jbxt)
~~~

3. open `jbxt.md`
4. call find-references on the language server client.

<details>
<summary>The links are recognized by `zk` as the graph output shows: </summary>
<pre>
$ zk graph --format=json | jq
{
  "notes": [
    {
      "filename": "eso7.md",
      "filenameStem": "eso7",
      "path": "old/eso7.md",
      "absPath": "/home/user/Documents/zk-test2/old/eso7.md",
      "title": "newly generated note",
      "link": "[newly generated note](old/eso7)",
      "lead": "generated with `zk new` directly in the old folder",
      "body": "generated with `zk new` directly in the old folder",
      "snippets": [
        "generated with `zk new` directly in the old folder"
      ],
      "rawContent": "# newly generated note\n\ngenerated with `zk new` directly in the old folder\n\n",
      "wordCount": 13,
      "tags": [],
      "metadata": {},
      "created": "2022-07-12T09:21:40.15040095Z",
      "modified": "2022-07-12T09:22:06.859619184Z",
      "checksum": "002388945a56d9be780edc458230055057439a5c89eb6474eb9f1490c906a1bb"
    },
    {
      "filename": "jbxt.md",
      "filenameStem": "jbxt",
      "path": "old/jbxt.md",
      "absPath": "/home/user/Documents/zk-test2/old/jbxt.md",
      "title": "ref-note",
      "link": "[ref-note](old/jbxt)",
      "lead": "Blub",
      "body": "Blub\n\nOriginal note: [top-note](vlix)",
      "snippets": [
        "Blub"
      ],
      "rawContent": "# ref-note\n\nBlub\n\nOriginal note: [top-note](vlix)\n",
      "wordCount": 6,
      "tags": [],
      "metadata": {},
      "created": "2022-07-12T09:16:54.491128191Z",
      "modified": "2022-07-12T09:18:30.498392496Z",
[...]

PR #246, opened 2022-07-14
LSP: Fix Double Use of notebook.RelPath

bibor:
Solves a bug introduced in PR #245. Discussion in #243.

notebook.RelPath() is called on a already relative path from the root. This causes a new, wrong relative path, because it is calculated from the note's paths.


mickael-menu (approved):
@bibor Thanks for spotting and fixing this!

Time to get started with automated tests on the LSP server...

mickael-menu:
I amended your commit without change, somehow GitHub actions were stuck.

PR #249, opened 2022-07-29
Link to mickael-menu/zk-nvim instead of megalithic/zk.nvim

kluen:
https://github.com/megalithic/zk.nvim has been archived for a while.

I'm assuming it is intended to link to the same plugin as in `README.md`.


mickael-menu (approved):
Yes, thank you for spotting that.

PR #262, opened 2022-09-26
Include get-date helper function

zalegrala:


internal/adapter/handlebars/helpers/getdate.go
```diff
@@ -0,0 +1,22 @@
...
+)
+
+// RegisterGetDate registers the {{getdate}} template helper to use the `naturaldate` package to generate time.Time based on language strings.
+// This can be used in combination with the `date` helper to generate dates in the user's language.
+// {{date (get-date "last week") "timestamp"}}
+func RegisterGetDate(logger util.Logger) {
+	raymond.RegisterHelper("get-date", func(natural string) time.Time {
+		date, err := naturaldate.Parse(natural, time.Now())
+		if err != nil {
+			logger.Printf("the {{get-date}} template helper failed to parse the date: %s", err)
```
mickael-menu, line 18: To properly wrap the error and report it, you can use this:

```suggestion
			logger.Err(errors.Wrap(err, "the {{get-date}} template helper failed to parse the date))
```

internal/adapter/handlebars/helpers/getdate.go
```diff
@@ -0,0 +1,22 @@
...
+
+	"github.com/aymerick/raymond"
+	"github.com/mickael-menu/zk/internal/util"
+	naturaldate "github.com/tj/go-naturaldate"
+)
+
+// RegisterGetDate registers the {{getdate}} template helper to use the `naturaldate` package to generate time.Time based on language strings.
+// This can be used in combination with the `date` helper to generate dates in the user's language.
+// {{date (get-date "last week") "timestamp"}}
+func RegisterGetDate(logger util.Logger) {
```
mickael-menu, line 14: You will need to register the helper here, otherwise it won't be available:

https://github.com/zalegrala/zk/blob/da5b7d12341016f937f35a9a9a84683af900a0e4/internal/adapter/handlebars/handlebars.go#L16

mickael-menu (changes requested):
Thank you! Would you mind adding some unit tests for this new helper? You can find examples here: https://github.com/mickael-menu/zk/blob/c21c4fc21f7fe75f05d3fc3d09e74ffed41ba2ef/internal/adapter/handlebars/handlebars_test.go#L229

zalegrala:
Nice, thanks for the review.  I'll find some time this week to add tests and address your feedback.

internal/adapter/handlebars/helpers/getdate.go
zalegrala: Updated, thank you.

zalegrala:
I've added a couple tests.  Let me know if you'd like me to drop the `date` calls in the `get-date` tests.  I added them because I mentioned them in the function docs, so thought it was worth leaving in.

internal/adapter/handlebars/helpers/getdate.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/pkg/errors"
+	naturaldate "github.com/tj/go-naturaldate"
+)
+
+// RegisterGetDate registers the {{getdate}} template helper to use the `naturaldate` package to generate time.Time based on language strings.
+// This can be used in combination with the `date` helper to generate dates in the user's language.
+// {{date (get-date "last week") "timestamp"}}
+func RegisterGetDate(logger util.Logger) {
+	raymond.RegisterHelper("get-date", func(natural string) time.Time {
+		date, err := naturaldate.Parse(natural, time.Now())
```
mickael-menu, line 17: Sorry I missed this in my first review, but you can use the helper [`TimeFromNatural()`](https://github.com/mickael-menu/zk/blob/c21c4fc21f7fe75f05d3fc3d09e74ffed41ba2ef/internal/util/date/date.go#L39) (which is part of this project) instead of using directly `naturaldate.Parse()`. This way it will be able to parse more formats.

mickael-menu (changes requested):
> I've added a couple tests. Let me know if you'd like me to drop the `date` calls in the `get-date` tests.
No that's great, thank you.

I think the tests are failing because the helper is using `time.Now()` in the implementation, so the `now` context var set in the tests is ignored.

To work around this, and after fixing my other comment to use `TimeFromNatural()`, you can use an absolute date as input (e.g. `{{get-date "2006-01-02T15:04:05"}}` instead of a relative one. We just need to test that `TimeFromNatural()` is called, and can assume that relative dates will work as well.

zalegrala:
Okay, I've updated.  As a side note, I had to add `exclude github.com/mattn/go-sqlite3 v2.0.3+incompatible` to the `go.mod` in order to test locally.  Looks like some chatter here about it.  https://github.com/mattn/go-sqlite3/issues/975

mickael-menu (approved):

mickael-menu:
@zalegrala Thank you for contributing this feature!

I might change the names before the next release, I'll sleep on it:
* `{{date}}` -> `{{format-date}}`
* `{{get-date}}` -> `{{date}}`

mickael-menu:
I released a new version and renamed the date helpers.

PR #266, opened 2022-11-01
Add exact-word match strategy

rktjmp:
**_Not sure if this is too niche to merge, so it's currently without tests/docs/etc but if you're interested in merging it upstream I will fill out the rest._**

I'm trying to recreate notational velocity's UI with `zk`.

Notation velocity has a very simple search interface, it just splits the query on whitespace and looks for each token inside a note via `strstr`. Searches are always `and`'d. 

`zk` can *almost* do this but:

- fts+stemming tends to return too broad results for some words, or too slim for partial tokens (eg: searching for `buf` wont find `buffer`, but searching for `logging` _will_ find `log`, you can fix `buf` with `buf*` but `"logging"` will still find `log`).
  - disabling stemming would be per database, AFAICT you can't set it per query, this option would be pretty invasive to the codebase.
- regular expressions can be more explicit but will only match tokens in the order given.
- exact match is actually pretty close, perfect for single tokens, but by design looks for "a b", not "a" and "b".

This adds a new match strategy `exact-word` which behaves similar to exact but splits the match filter into words first.

Not sure if "exact-**word**" really describes the behaviour, as `dog` will intentionally match `dogbowl`, `catdog` and `dog`, -- so it's really exact-strings? substrings?  exact-terms? exact-word**s**? `--match-strategy notational-velocity`?

Alternatively, instead of adding an new strategy, the `-m` option could be extended to support multiple instances: `-m word -m word2`.

### DIFF
--- a/internal/adapter/sqlite/note_dao.go
+++ b/internal/adapter/sqlite/note_dao.go
@@ -524,6 +524,14 @@ func (d *NoteDAO) findRows(opts core.NoteFindOpts, selection noteSelection) (*sq
 		case core.MatchStrategyExact:
 			whereExprs = append(whereExprs, `n.raw_content LIKE '%' || ? || '%' ESCAPE '\'`)
 			args = append(args, escapeLikeTerm(opts.Match.String(), '\\'))
+		case core.MatchStrategyExactWord:
+			words := strings.Fields(opts.Match.String())
+			for range words {
+				whereExprs = append(whereExprs, `n.raw_content LIKE '%' || ? || '%' ESCAPE '\'`)
+			}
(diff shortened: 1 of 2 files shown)


mickael-menu:
Thank you for your contribution @rktjmp, this is a great use case.

I think it might be a bit too specific for a match strategy. However, the multiple `-m` options could be a great addition and be more versatile. Filtering options are usually combined with `AND` so it makes sense.

Are you open to taking a stab at the implementation? Let me know if I can guide you through the codebase.

rktjmp:
Sounds good, opened #268

PR #267, opened 2022-11-04
Fix --working-dir= and --notebook-dir= option parsing

rktjmp:
--working-dir and --notebook-dir options run through some custom parsing code before everything else so it can apply to aliases.

When given in the form --working-dir=x (with an =), the previous implementation would check this as a single option instead of separating it into an option=value pair, which would mean --working-dir=, -W= and --notebook-dir= would not be recognised when given.

### DIFF
--- a/main.go
+++ b/main.go
@@ -7,6 +7,7 @@ import (
 	"os/exec"
 	"os/signal"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"strings"
 
@@ -204,9 +205,9 @@ func runAlias(container *cli.Container, args []string) (bool, error) {
 // which path arguments are relative from.
 //
 // By order of precedence:
-//   1. --notebook-dir flag
-//   2. current working directory
-//   3. ZK_NOTEBOOK_DIR environment variable
+//  1. --notebook-dir flag
+//  2. current working directory
+//  3. ZK_NOTEBOOK_DIR environment variable
 func notebookSearchDirs(dirs cli.Dirs) ([]cli.Dirs, error) {
 	wd, err := os.Getwd()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -204,9 +205,9 @@ func runAlias(container *cli.Container, args []string) (bool, error) {
 // which path arguments are relative from.
 //
 // By order of precedence:
-//   1. --notebook-dir flag
-//   2. current working directory
-//   3. ZK_NOTEBOOK_DIR environment variable
+//  1. --notebook-dir flag
+//  2. current working directory
+//  3. ZK_NOTEBOOK_DIR environment variable
```
rktjmp, line 210: `gofmt` changed this.

rktjmp:
Sorry, there was deficiency where `zk --notebook-dir=` would not print an error to the user.

I tried to just drop Kong in but Kong Scanner was too low level (doesn't actually split `=`) and I could not figure how to convince the parser to not give up on unknown values - probably why you went the way you did originally, so re-opening this with a modified version of the first that fixes the bug and consolidates some of the code paths to hopefully make it clearer.

rktjmp:
Just remembered this wont show nice errors if `--notebook-dir` is the last option, it does give an error though.

`zk: error: --notebook-dir: expected file value but got "EOL" (<EOL>)`

main.go
mickael-menu: That's fine.

mickael-menu (approved):
Thank you for the fix, looking good 👍 

> I tried to just drop Kong in but Kong Scanner was too low level (doesn't actually split =) and I could not figure how to convince the parser to not give up on unknown values and go's flag lib wont correctly our short vs long flags - probably why you went the way you did originally, so re-opening this with a modified version of the first that fixes the bug and consolidates some of the code paths to hopefully make it clearer.

I don't remember exactly but I initially wanted to use Kong twice but couldn't make it work, hence the manual parsing.

PR #268, opened 2022-11-06
Mmmmulti match

rktjmp:
Alternative to #266

- [x] update at least one `docs/` file
- [x] update all `zk help` output
- [x] confirm `copy()` requirements?

### DIFF
--- a/internal/adapter/handlebars/handlebars_test.go
+++ b/internal/adapter/handlebars/handlebars_test.go
@@ -249,12 +249,12 @@ func TestDateHelper(t *testing.T) {
 	testString(t, "{{date now 'timestamp'}}", context, "200911172034")
 	testString(t, "{{date now 'timestamp-unix'}}", context, "1258490098")
 	testString(t, "{{date now 'cust: %Y-%m'}}", context, "cust: 2009-11")
-	testString(t, "{{date now 'elapsed'}}", context, "13 years ago")
+	testStringMatch(t, "{{date now 'elapsed'}}", context, `[\d]+ years ago`)
 }
 
 func TestGetDateHelper(t *testing.T) {
 	context := map[string]interface{}{"now": time.Date(2009, 11, 17, 20, 34, 58, 651387237, time.UTC)}
-	testString(t, "{{get-date \"2009-11-17T20:34:58\"}}", context, "2009-11-17 20:34:58 +0000 UTC")
+	testStringMatch(t, "{{get-date \"2009-11-17T20:34:58\"}}", context, `2009-11-17 20:34:58 \+[\d]{4} [A-Z]+`)
 }
 
 func TestShellHelper(t *testing.T) {
--- a/internal/adapter/handlebars/handlebars_test.go
+++ b/internal/adapter/handlebars/handlebars_test.go
@@ -45,17 +45,6 @@ func testString(t *testing.T, template string, context interface{}, expected str
 	assert.Equal(t, actual, expected)
 }
 
-func testStringMatch(t *testing.T, template string, context interface{}, pattern string) {
-	sut := testLoader(LoaderOpts{})
-
-	templ, err := sut.LoadTemplate(template)
-	assert.Nil(t, err)
-
-	actual, err := templ.Render(context)
-	assert.Nil(t, err)
-	assert.Regexp(t, actual, pattern)
-}
-
 func testFile(t *testing.T, name string, context interface{}, expected string) {
 	sut := testLoader(LoaderOpts{})
 
@@ -249,12 +238,13 @@ func TestDateHelper(t *testing.T) {
 	testString(t, "{{date now 'timestamp'}}", context, "200911172034")
 	testString(t, "{{date now 'timestamp-unix'}}", context, "1258490098")
 	testString(t, "{{date now 'cust: %Y-%m'}}", context, "cust: 2009-11")
-	testStringMatch(t, "{{date now 'elapsed'}}", context, `[\d]+ years ago`)
+	testString(t, "{{date now 'elapsed'}}", context, "14 years ago")
 }
 
 func TestGetDateHelper(t *testing.T) {
 	context := map[string]interface{}{"now": time.Date(2009, 11, 17, 20, 34, 58, 651387237, time.UTC)}
-	testStringMatch(t, "{{get-date \"2009-11-17T20:34:58\"}}", context, `2009-11-17 20:34:58 \+[\d]{4} [A-Z]+`)
+	localOffsetAndTZ := time.Now().Format("-0700 MST")
+	testString(t, "{{get-date \"2009-11-17T20:34:58\"}}", context, "2009-11-17 20:34:58 "+localOffsetAndTZ)
 }
 
 func TestShellHelper(t *testing.T) {
--- a/internal/adapter/handlebars/handlebars_test.go
+++ b/internal/adapter/handlebars/handlebars_test.go
@@ -45,6 +45,17 @@ func testString(t *testing.T, template string, context interface{}, expected str
 	assert.Equal(t, actual, expected)
 }
 
+func testStringMatch(t *testing.T, template string, context interface{}, pattern string) {
+	sut := testLoader(LoaderOpts{})
+
+	templ, err := sut.LoadTemplate(template)
+	assert.Nil(t, err)
+
+	actual, err := templ.Render(context)
+	assert.Nil(t, err)
+	assert.Regexp(t, actual, pattern)
+}
+
 func testFile(t *testing.T, name string, context interface{}, expected string) {
 	sut := testLoader(LoaderOpts{})
 
--- a/internal/adapter/sqlite/note_dao.go
+++ b/internal/adapter/sqlite/note_dao.go
@@ -420,10 +420,7 @@ func (d *NoteDAO) expandMentionsIntoMatch(opts core.NoteFindOpts) (core.NoteFind
 	}
 
 	// Expand the mention queries in the match predicate.
-	match := make([]string, len(opts.Match))
-	copy(match, opts.Match)
-	match = append(match, " ("+strings.Join(mentionQueries, " OR ")+") ")
-	opts.Match = match
+	opts.Match = append(opts.Match, " ("+strings.Join(mentionQueries, " OR ")+") ")
 
 	return opts, nil
 }
--- a/internal/adapter/sqlite/note_dao.go
+++ b/internal/adapter/sqlite/note_dao.go
@@ -12,7 +12,6 @@ import (
 	"github.com/mickael-menu/zk/internal/util"
 	"github.com/mickael-menu/zk/internal/util/errors"
 	"github.com/mickael-menu/zk/internal/util/fts5"
-	"github.com/mickael-menu/zk/internal/util/opt"
 	"github.com/mickael-menu/zk/internal/util/paths"
 	strutil "github.com/mickael-menu/zk/internal/util/strings"
 )
@@ -421,9 +420,10 @@ func (d *NoteDAO) expandMentionsIntoMatch(opts core.NoteFindOpts) (core.NoteFind
 	}
 
 	// Expand the mention queries in the match predicate.
-	match := opts.Match.String()
-	match += " " + strings.Join(mentionQueries, " OR ")
-	opts.Match = opt.NewString(match)
+	match := make([]string, len(opts.Match))
+	copy(match, opts.Match)
+	match = append(match, " ("+strings.Join(mentionQueries, " OR ")+") ")
+	opts.Match = match
 
 	return opts, nil
 }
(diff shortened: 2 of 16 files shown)


internal/adapter/sqlite/note_dao.go
```diff
@@ -421,9 +420,10 @@ func (d *NoteDAO) expandMentionsIntoMatch(opts core.NoteFindOpts) (core.NoteFind
 	}
 
 	// Expand the mention queries in the match predicate.
-	match := opts.Match.String()
-	match += " " + strings.Join(mentionQueries, " OR ")
-	opts.Match = opt.NewString(match)
+	match := make([]string, len(opts.Match))
+	copy(match, opts.Match)
+	match = append(match, " ("+strings.Join(mentionQueries, " OR ")+") ")
+	opts.Match = match
```
rktjmp, line 426: Not codebase-fluent (or go-fluent) enough to know whether we should copy here to avoid altering `x.Match` in other terms. Sort of just wrote it with a mind to check later.

internal/adapter/sqlite/note_dao.go
```diff
@@ -519,20 +519,26 @@ func (d *NoteDAO) findRows(opts core.NoteFindOpts, selection noteSelection) (*sq
...
+				whereExprs = append(whereExprs, "fts_match.notes_fts MATCH ?")
+				args = append(args, fts5.ConvertQuery(match))
+			}
 		case core.MatchStrategyRe:
-			whereExprs = append(whereExprs, "n.raw_content REGEXP ?")
-			args = append(args, opts.Match.String())
+			for _, match := range opts.Match {
+				whereExprs = append(whereExprs, "n.raw_content REGEXP ?")
+				args = append(args, match)
+			}
```
rktjmp, line 541: Assume we want to apply the behaviour to all strategies, all AND

internal/adapter/sqlite/note_dao_test.go
```diff
@@ -452,10 +452,25 @@ func TestNoteDAOFindMatch(t *testing.T) {
 	)
 }
 
+func TestNoteDAOFindMatchWithMultiMatch(t *testing.T) {
+	testNoteDAOFindPaths(t,
+		core.NoteFindOpts{
+			Match:         []string{"daily | index", "second"},
+			MatchStrategy: core.MatchStrategyFts,
```
rktjmp, line 459: Did not test all strats here (do in tesh tests), but these seemed brief anyway (e.g. does not currently test stemming, etc). Didn't want to boat the file but can add at request.

internal/cli/filtering.go
```diff
@@ -19,7 +18,7 @@ type Filtering struct {
 
 	Interactive    bool     `kong:"group='filter',short='i',help='Select notes interactively with fzf.'" json:"-"`
 	Limit          int      `kong:"group='filter',short='n',placeholder='COUNT',help='Limit the number of notes found.'" json:"limit"`
-	Match          string   `kong:"group='filter',short='m',placeholder='QUERY',help='Terms to search for in the notes.'" json:"match"`
```
rktjmp, line 22: Language pref?

> may be given multiple times

> multiple occurrences (instances?) are joined with an implicit AND

?

internal/cli/filtering.go
```diff
@@ -117,10 +116,13 @@ func (f Filtering) ExpandNamedFilters(filters map[string]string, expandedFilters
 				f.ModifiedAfter = parsedFilter.ModifiedAfter
 			}
 
-			if f.Match == "" {
+			if len(f.Match) == 0 {
 				f.Match = parsedFilter.Match
-			} else if parsedFilter.Match != "" {
-				f.Match = fmt.Sprintf("(%s) AND (%s)", f.Match, parsedFilter.Match)
+			} else if 0 < len(parsedFilter.Match) {
+				// f.Match = append(f.Match, fmt.Sprintf("(%s) AND (%s)", f.Match, parsedFilter.Match))
```
rktjmp, line 122: TODO: remove

internal/cli/filtering.go
```diff
@@ -148,7 +150,8 @@ func (f Filtering) NewNoteFindOpts(notebook *core.Notebook) (core.NoteFindOpts,
 		return opts, fmt.Errorf("the --exact-match (-e) option is deprecated, use --match-strategy=exact (-Me) instead")
 	}
 
-	opts.Match = opt.NewNotEmptyString(f.Match)
+	opts.Match = make([]string, len(f.Match))
+	copy(opts.Match, f.Match)
```
rktjmp, line 154: Again, safety copy? Needed?

internal/cli/filtering_test.go
```diff
@@ -168,7 +168,7 @@ func TestExpandNamedFiltersJoinMatch(t *testing.T) {
 	)
 
 	assert.Nil(t, err)
-	assert.Equal(t, res.Match, "(((chocolate OR caramel)) AND (banana)) AND (apple)")
+	assert.Equal(t, res.Match, []string{"(chocolate OR caramel)", "banana", "apple"})
```
rktjmp, line 171: Implicit AND now, so had to alter test case expectation

internal/core/note_find.go
```diff
@@ -5,14 +5,12 @@ import (
...
 	"unicode/utf8"
-
-	"github.com/mickael-menu/zk/internal/util/opt"
 )
 
 // NoteFindOpts holds a set of filtering options used to find notes.
 type NoteFindOpts struct {
 	// Filter used to match the notes with the given MatchStrategy.
-	Match opt.String
+	Match []string
```
rktjmp, line 13: Dropped the opt string type here and in filter since the collection operates in a similar enough manner (len = 0 == nullString), did not see reason for double encapsulation.

internal/core/notebook.go
```diff
@@ -236,9 +236,9 @@ func (n *Notebook) FindByHref(href string, allowPartialHref bool) (*MinimalNote,
 }
 
 // FindMatching retrieves the first note matching the given search terms.
-func (n *Notebook) FindMatching(terms string) (*MinimalNote, error) {
+func (n *Notebook) FindMatching(terms []string) (*MinimalNote, error) {
 	return n.FindMinimalNote(NoteFindOpts{
-		Match: opt.NewNotEmptyString(terms),
+		Match: terms,
 	})
```
rktjmp, line 242: btw this is never called anywhere.

tests/cmd-list-filter-match-exact.tesh
```diff
@@ -15,3 +15,34 @@ $ zk list -q --debug-style -Me --match '["न", "म", "स्", "ते"]'
 >
 >  - Given the Hindi word "नमस्ते":
 >
+
+# Mutliple match flags.
+$ zk list -q --debug-style -Me --match "thread" --match "mut"
+><title>Concurrency in Rust</title> <path>g7qa.md</path> (9 minutes ago)
```
rktjmp, line 21: The timestamps here were giving me some grief, I ran `make tesh-update` which I guess is the idea? It was pretty violent though so I only included the "new" tests but I assume the timestamps need to be manually changed or ...

internal/adapter/sqlite/note_dao.go
rktjmp: Actually don't think it's need as append does its own copy.

internal/cli/filtering.go
rktjmp: I think this one should be a copy as `opts.Match = f.Match` would assign the `[]string{}` pointers - because slices are structs - and so changes could leak upwards?

tests/cmd-list-filter-match-exact.tesh
rktjmp: I think this date issue might have been because I had a `notebook.db` in tree, removing it fixes it as the index is always fresh.

mickael-menu: Yeah that's right, the database is rebuilt from scratch for every tesh file.

internal/core/notebook.go
mickael-menu: Ha that's probably some leftover from a previous refactoring. Feel free to remove it if you want.

internal/core/note_find.go
mickael-menu: Absolutely, that's the right call.

internal/cli/filtering_test.go
mickael-menu: Looks good.

internal/adapter/sqlite/note_dao.go
mickael-menu: Yes

internal/cli/filtering.go
```diff
@@ -19,7 +18,7 @@ type Filtering struct {
 
 	Interactive    bool     `kong:"group='filter',short='i',help='Select notes interactively with fzf.'" json:"-"`
 	Limit          int      `kong:"group='filter',short='n',placeholder='COUNT',help='Limit the number of notes found.'" json:"limit"`
-	Match          string   `kong:"group='filter',short='m',placeholder='QUERY',help='Terms to search for in the notes.'" json:"match"`
+	Match          []string `kong:"group='filter',short='m',placeholder='QUERY',help='Terms to search for in the notes. If given multiple times each argument is joined with a boolean AND.'" json:"match"`
```
mickael-menu, line 21: This is true for most other options, so I don't think it's necessary to mention it. It's also described by Kong in the flag name:

```
-m, --match=QUERY,...
```
```suggestion
	Match          []string `kong:"group='filter',short='m',placeholder='QUERY',help='Terms to search for in the notes."`
```

internal/cli/filtering.go
```diff
@@ -117,10 +116,12 @@ func (f Filtering) ExpandNamedFilters(filters map[string]string, expandedFilters
...
 
-			if f.Match == "" {
+			if len(f.Match) == 0 {
 				f.Match = parsedFilter.Match
-			} else if parsedFilter.Match != "" {
-				f.Match = fmt.Sprintf("(%s) AND (%s)", f.Match, parsedFilter.Match)
+			} else if 0 < len(parsedFilter.Match) {
+				for _, pmatch := range parsedFilter.Match {
+					f.Match = append(f.Match, pmatch)
+				}
```
mickael-menu, line 124: Now that we use slices, this can all be simplified to:

```go
f.Match = append(f.Match, parsedFilter.Match...)
```

`append` is a variadic function and [in Go you can use the `...` spread operator](https://go.dev/ref/spec#Passing_arguments_to_..._parameters) to pass each element of a slice to a variadic function.

internal/cli/filtering.go
mickael-menu: Right, any change in `opts.Match` would impact `f.Match`.

mickael-menu (changes requested):
Thank you for submitting this contribution and thinking of updating the docs @rktjmp! I left a few comments but nothing major.

internal/adapter/sqlite/note_dao.go
```diff
@@ -519,20 +516,26 @@ func (d *NoteDAO) findRows(opts core.NoteFindOpts, selection noteSelection) (*sq
 		return nil
 	}
 
-	if !opts.Match.IsNull() {
+	if 0 < len(opts.Match) {
```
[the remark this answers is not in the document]
rktjmp: Doing that breaks seems to break `case core.MatchStrategyFts:` as we add duplicate `joinClauses`, probably cleaner to keep it all local rather than pop the fts query construction out separately?

```
    assert.go:28: Expected `ambiguous column name: fts_match.notes_fts` (type sqlite3.Error) to be nil
    assert.go:45: Received (type []string):
        []string{}
    assert.go:46:
        ---
    assert.go:47: But expected (type []string):
        []string{"log/2021-01-04.md"}
    assert.go:48:
        ---
    assert.go:49: Diff:
    assert.go:51:       "[]string[0] != []string[1]"
````

**Comment by rktjmp** [comment]
Thanks for taking the time to review and answer my questions! I replied to some  :arrow_up:.

File: internal/adapter/handlebars/handlebars_test.go [hunk 14]
```diff
@@ -238,12 +238,13 @@ func TestDateHelper(t *testing.T) {
...
 	testString(t, "{{date now 'cust: %Y-%m'}}", context, "cust: 2009-11")
-	testString(t, "{{date now 'elapsed'}}", context, "13 years ago")
+	testString(t, "{{date now 'elapsed'}}", context, "14 years ago")
 }
 
 func TestGetDateHelper(t *testing.T) {
 	context := map[string]interface{}{"now": time.Date(2009, 11, 17, 20, 34, 58, 651387237, time.UTC)}
-	testString(t, "{{get-date \"2009-11-17T20:34:58\"}}", context, "2009-11-17 20:34:58 +0000 UTC")
+	localOffsetAndTZ := time.Now().Format("-0700 MST")
+	testString(t, "{{get-date \"2009-11-17T20:34:58\"}}", context, "2009-11-17 20:34:58 "+localOffsetAndTZ)
```
rktjmp, line 247: Test would fail because the string included my local tz and offset, assume this wasn't regularly caught because CI is always in UTC and perhaps you run it locally in a container or whatever. I did turn it into a regexp match in 5574400a0ff7fa77ad0bad8781b3ddd4473ec928 but reverted back to an explicit match as the intention is clearer.

internal/adapter/sqlite/note_dao.go
```diff
@@ -421,9 +420,7 @@ func (d *NoteDAO) expandMentionsIntoMatch(opts core.NoteFindOpts) (core.NoteFind
 	}
 
 	// Expand the mention queries in the match predicate.
-	match := opts.Match.String()
-	match += " " + strings.Join(mentionQueries, " OR ")
-	opts.Match = opt.NewString(match)
+	opts.Match = append(opts.Match, " ("+strings.Join(mentionQueries, " OR ")+") ")
```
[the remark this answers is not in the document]
mickael-menu: You're right, thanks for clarifying 👍

internal/adapter/handlebars/handlebars_test.go
mickael-menu: Sounds good

internal/adapter/sqlite/note_dao.go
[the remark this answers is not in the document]
mickael-menu: Ha yes, I missed that.

mickael-menu (approved):
Thanks a lot for taking the time to contribute this feature and addressing the comments! 🙏  I will release a new version with it soon.

PR #281, opened 2023-01-04
Add zk-spaced to list of related projects

matze:
Also take the opportunity to unify the dashes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,5 +94,6 @@ $ ./zk -h
 ## Related projects
 
 * [Neuron](https://github.com/srid/neuron) – a great tool to publish a Zettelkasten on the web
-* [Emanote](https://emanote.srid.ca/) - an improved successor to Neuron
+* [Emanote](https://emanote.srid.ca/) – an improved successor to Neuron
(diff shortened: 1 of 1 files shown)


mickael-menu (approved):
Thank you, that looks like an interesting project!

I thought of adding SRS features in `zk` directly, but as an external plugin it makes much sense. Let me know if something could be changed in `zk` to help you build `zk-spaced`.

matze:
> Let me know if something could be changed in zk to help you build zk-spaced.

I think negotiating the output format without having the user do it would be nice. But I don't think it's worth making the whole interaction more complex. So, all good :-)

mickael-menu:
Having `zk-*` binaries implicitly available as `zk` subcommands could be a minimal way to support external plugins seamlessly. For example your hypothetical `zk-review` binary would call `zk` directly with the expected arguments (`--format json`) and it would be available to the user as `zk review` without modifying their config.

I think it's pretty standard, see for example `git-lfs`.

matze:
Sounds like a reasonable approach, yes.

PR #283, opened 2023-01-07
Fix emanote URL

leiserfg:


mickael-menu (approved):
Thank you for spotting that!

PR #284, opened 2023-01-10
feat(lsp): add 'zk.link' LSP command

psanker:
Allows users to take advantage of the linking mechanism in their editing tools outside of the 'zk.new' context. For example, I made a "ZkInsertLink" command with Neovim that inserts a link at the current cursor location after searching for the desired note. This improves on the current behavior which only autocompletes based on note name, which users may or may not remember.

I looked for a `CONTRIBUTING.md` but couldn't find. I hope you find this PR useful!

## Changes
- Adds 'zk.link' LSP command
- Refactors linking mechanism from `executeCommandNew` into utility function to be used by both `executeCommandNew` and `executeCommandLink`

### DIFF
--- a/internal/adapter/lsp/cmd_link.go
+++ b/internal/adapter/lsp/cmd_link.go
@@ -38,7 +38,11 @@ func executeCommandLink(notebook *core.Notebook, documents *documentStore, conte
 
 	note, err := notebook.FindByHref(*opts.Path, false)
 
-	if err != nil || note == nil {
+	if err != nil {
+		return nil, err
+	}
+
+	if note == nil {
 		return nil, errors.New("Requested note to link to not found!")
 	}
 
@@ -48,14 +52,13 @@ func executeCommandLink(notebook *core.Notebook, documents *documentStore, conte
 		title:    opts.Title,
 	}
 
-    ok, err := linkNote(notebook, documents, context, info)
+    err = linkNote(notebook, documents, context, info)
 
-    if !ok {
+    if err != nil {
         return nil, err
     }
 
 	return map[string]interface{}{
-		"ok":   true,
 		"path": filepath.Join(notebook.Path, note.Path),
 	}, nil
 }
--- a/internal/adapter/lsp/cmd_link.go
+++ b/internal/adapter/lsp/cmd_link.go
@@ -1,0 +1,61 @@
+package lsp
+
+import (
+	"fmt"
+	"path/filepath"
+
+	"github.com/mickael-menu/zk/internal/core"
+	"github.com/mickael-menu/zk/internal/util/errors"
+	"github.com/tliron/glsp"
+	protocol "github.com/tliron/glsp/protocol_3_16"
+)
+
+const cmdLink = "zk.link"
+
+type cmdLinkOpts struct {
+	Path     *string            `json:"path"`
+	Location *protocol.Location `json:"location"`
+	Title    *string            `json:"title"`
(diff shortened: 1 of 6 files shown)


internal/adapter/lsp/cmd_link.go
```diff
@@ -0,0 +1,61 @@
...
+
+	if opts.Path == nil {
+		return nil, errors.New("'path' not provided")
+	}
+
+	note, err := notebook.FindByHref(*opts.Path, false)
+
+	if err != nil || note == nil {
+		return nil, errors.New("Requested note to link to not found!")
+	}
```
mickael-menu, line 43: The error returned by `FindByHref` might be of interest.

```suggestion
	if err != nil {
	    return nil, err
	}
	if note == nil {
		return nil, errors.New("target note not found")
	}
```

internal/adapter/lsp/cmd_link.go
```diff
@@ -0,0 +1,61 @@
...
+
+    ok, err := linkNote(notebook, documents, context, info)
+
+    if !ok {
+        return nil, err
+    }
+
+	return map[string]interface{}{
+		"ok":   true,
```
mickael-menu, line 58: I'm not sure about adding this `ok` property as it will always be `true`. What did you have in mind for its usage?

internal/adapter/lsp/util.go
```diff
@@ -56,3 +60,59 @@ func (b *jsonBoolean) UnmarshalJSON(data []byte) error {
...
 	return nil
 }
+
+type linkInfo struct {
+    note *core.MinimalNote
+	location *protocol.Location
+	title    *string
+}
+
+func linkNote(notebook *core.Notebook, documents *documentStore, context *glsp.Context, info *linkInfo) (bool, error) {
```
mickael-menu, line 70: You can simplify this a bit by returning a single `error`, as `bool` is `true` only when the error is `nil`.

mickael-menu (commented):
Thank you for your contribution, this will be very useful!

Would you mind [updating the documentation](https://github.com/mickael-menu/zk/blob/main/docs/editors-integration.md#custom-commands) to add the new `zk.link` command?

mickael-menu:
Could you also `git apply` the following patch to update the changelog? I don't have the rights to push on your fork directly. (If you open another PR, you can check the "allow edits from maintainers" checkbox when creating the PR). Thanks!

[changelog.patch](https://github.com/mickael-menu/zk/files/10438953/changelog.patch)

psanker:
Done! Added your suggestions and the changelog patch. As I note in the commits, the `bool` values were vestigial values, originally there if there was extra information that could have been useful to send back. Turns out there wasn't really.

mickael-menu:
Looks like the build fails with the recent changes.

```
GOARCH=arm64 go build -tags "fts5" -ldflags "-X=main.Version=`git describe --tags --match v[0-9]* 2> /dev/null` -X=main.Build=`git rev-parse --short HEAD`"
# github.com/mickael-menu/zk/internal/adapter/lsp
internal/adapter/lsp/cmd_new.go:93:20: assignment mismatch: 2 variables but linkNote returns 1 value
make: *** [build] Error 2
```

> Would you be interested in a PR for these on zk-nvim?

Sure! That would be useful. Thanks again.

psanker:
Whoops! Thought I caught all of those refactors. `make install` now passes cleanly for me.

PR #288, opened 2023-02-06
Adds "title" to the Command in code actions

dutrieuc:
This fixes issue with some clients (see for example this [kak-lsp issue](https://github.com/kak-lsp/kak-lsp/issues/669)) expecting to find a non empty string there to display.


mickael-menu (approved):
Looking good, thank you 👍

PR #302, opened 2023-03-24
Add `shell` to configuration file

lsvmello:
Closes https://github.com/mickael-menu/zk/issues/300

I didn't added the logic to windows because it gets a little more complicated there. Most shell on linux uses the `-c` flag but on windows the flag might be completely different: `/c` for _cmd_, `-NoInteractive` for _powershell_ and `-c` for _zsh_, _bash_ and _fish_.


internal/util/os/os.go
```diff
@@ -10,7 +10,10 @@ import (
 // Getenv returns an optional String for the environment variable with given
 // key.
 func GetOptEnv(key string) opt.String {
-	return opt.NewNotEmptyString(os.Getenv(key))
```
mickael-menu, line 13: Isn't this doing the same thing? `Getenv` will return `""` if the key is missing, which will be converted to `opt.NullString` by `NewNotEmptyString()`.

docs/config.md
```diff
@@ -63,10 +64,10 @@ author = "Mickaël"
 
 
 # GROUP OVERRIDES
-[dir.journal]
```
mickael-menu, line 66: Ha good catch!

mickael-menu (commented):
Thank you for tackling this, and in particular for filling the documentation 👍

internal/util/os/os.go
lsvmello: Unfortunately no, after a few tests I ended up using the recommendation on the [docs](https://pkg.go.dev/os#Getenv).

> Getenv retrieves the value of the environment variable named by the key. It returns the value, which will be empty if the variable is not present. ***To distinguish between an empty value and an unset value, use LookupEnv.***

So if you call `Getenv` with an unset environment variable the value returned from that function is not `""`.

I found the `LookupEnv` in other parts of the code, so I figured you had already bumped into this.

mickael-menu (approved):
Thanks again @lsvmello!

PR #317, opened 2023-05-12
Use UTF-16 to get the correct position

wrvsrx:
Fix https://github.com/mickael-menu/zk-nvim/issues/69.

According to https://pkg.go.dev/go.lsp.dev/protocol#Position, "The offsets are based on a UTF-16 string representation".

Other code that uses protocol.Position may also have similar issues, but I only fixed the part that affects autocompletion.

### DIFF
--- a/internal/adapter/lsp/document.go
+++ b/internal/adapter/lsp/document.go
@@ -5,6 +5,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
+	"unicode/utf16"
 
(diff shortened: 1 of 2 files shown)


mickael-menu (approved):
`zk` is using [tliron/glsp](github.com/tliron/glsp), but according to the [LSP specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/) UTF-16 offsets are the default, unless specified otherwise. So I guess that applies too.

Thanks for taking the time to send a fix.

PR #354, opened 2023-11-07
Address io/ioutil deprecation

julio-lopez:
Package io/ioutil was deprecated in Go 1.16.

This replaces the use of functions from io/ioutil with equivalent ones from the io and os packages. This allows to more easily upgrade the version of Go used to build zk.

Refs:
  https://pkg.go.dev/io/ioutil


mickael-menu (approved):
Thank you Julio 🙏

PR #372, opened 2024-01-03
Update the arch repo link

mcDevnagh:


tjex (approved):
yep, link checks out. Upstream url in the repo itself will also have to be changed of course. Still points to ../mickael-menu/zk

PR #374, opened 2024-01-03
Relative output needs relative input

mcDevnagh:


tjex (approved):
All makes sense.
Also ran the test in question on my end and it passes.

PR #376, opened 2024-01-05
notification of maintenance mode

tjex:
To keep all users up to date with the change, and clear up any confusion about
the url redirect they get when visiting any repos that were at github.com/mickael-menu/...


mickael-menu (approved):
Thanks Tillman, this is elegantly put!

PR #381, opened 2024-01-13
Fix broken links

mcDevnagh:
Missed these in #373


tjex (approved):
truuuuuuu! good catch

Issue #382, opened 2024-01-17
zk list --created date, returns notes from one day after if YAML date does not have timestamp


tjex:
### Check if applicable

- [X] I have searched the existing issues (**required**)
- [X] I'm willing to help fix the problem and contribute a pull request

### Describe the bug

I created a daily note today (17th Jan). The system metadata via Finder shows that the file was created today, and the template for naming my daily notes uses `{{format-date now}}`, yet when I run `zk list --created=2024-01-17` the note is not returned.

It is returned if I run `zk list --created=2024-17-16`

```bash
$ zk list --created=2024-01-16
2024-01-17 d/2024/2024-01-17.md (16 hours ago)
```
Proof of date creation via `ll -l`

```bash
.rw-r--r-- tjex staff 935B Wed 17 Jan 2024-01-17.md
```

### How to reproduce?

1. Setup daily note conf as instructed [here](https://github.com/zk-org/zk/blob/main/docs/daily-journal.md)
2. create daily note and save
3. run `zk list --created=<the date today in YYYY-MM-DD format>
4. see if the daily note is returned or not

### zk configuration

```toml
[group.daily]
paths = ["d/2024"]

[group.daily.note]
filename = "{{format-date now}}"
extension = "md"

[alias]
daily = 'zk new --no-input "$ZK_NOTEBOOK_DIR/d/2024"'
```

### Environment

```bash
zk 0.14.0
system: Darwin 22.6.0 arm64
```

tjex:
Well... Its strange. If I replicate the steps in my testing vault, it works as expected. Will need to investigate (much) further.

tjex:
Getting this on `zk new` 

`zk: warning: sql: Scan error on column index 2, name "source_path": converting NULL to string is unsupported`

But again, not in the test vault, so this will be something in my config I'm guessing

tjex:
Deleted and regenerated my notebook.md. This fixed the error message. But whenever I create a new note, and print out the data on it, it states it was created 17 hours ago (the same as in the original case of my daily note)

tjex:
If I copy the file from my own vault, to the testing vault, and run `zk index`, the file that was created today *also* reports that it was created (now) 17 hours ago.

This is despite the 'bug' not being present in the testing vault whereby I can create a note with `zk new` and it will show up in `zk list --created=2024-01-17`.

If I run the same process in my own vault, the note will not be returned under notes created on 2024-01-17.

This is super weird...

tjex:
[...]

PR #384, opened 2024-01-22
fix day range parsing (zk-org/zk#382)

tjex:
fix: set day range end to be one second before midnight.

i.e, the next day begins at 00:00.

Now the day range extends to 23:59:59. Meaning that the notes created within the
same day, that do not have time stamps, will be returned by a query such as `zk
list --created=today`, previously would fall under `... --created=yesterday`.

Likewise, given a note with the YAML date `2024-01-22`, zk would report this as being
created on `2024-01-21`.

### DIFF
--- a/internal/cli/filtering.go
+++ b/internal/cli/filtering.go
@@ -280,7 +280,9 @@ func parseDayRange(date string) (start time.Time, end time.Time, err error) {
 		return
 	}
 
-	start = startOfDay(day)
+    // we add -1 second so that the day range ends at 23:59:59
+    // i.e, the 'new day' begins at 00:00:00
+	start = startOfDay(day).Add(time.Second * -1)
 	end = start.AddDate(0, 0, 1)
 	return start, end, nil
 }


jurica (approved):
Sounds sensible and change LGTM.

mcDevnagh:
Could we get a test for this fix, so we don't see a regression?

Like the following: https://github.com/zk-org/zk/blob/537c7f7554a90572873b46d88aaac4490d113d5f/internal/adapter/handlebars/handlebars_test.go#L243 but in this file: https://github.com/zk-org/zk/blob/main/internal/cli/filtering_test.go

It would need to fail before your changes, and succeed with the'

tjex:
yesss. good idea 😅
Will do so.

internal/cli/filtering.go
```diff
@@ -280,7 +280,9 @@ func parseDayRange(date string) (start time.Time, end time.Time, err error) {
 		return
 	}
 
-	start = startOfDay(day)
+    // we add -1 second so that the day range ends at 23:59:59
+    // i.e, the 'new day' begins at 00:00:00
+	start = startOfDay(day).Add(time.Second * -1)
```
julio-lopez, line 285: What was the value of start before? `00:00:00` or `00:00:01` ?
If start was already `00:00:00` then it seems that subtracting a second should be done to from `end`.

tjex: Here's the test: https://github.com/zk-org/zk/pull/385

PR #385, opened 2024-01-24
tesh test case for yaml dates without time stamp

tjex:
This adds a test case for zk-org/zk#382 which found that dates without time
stamps in yaml frontmatter were causing the respective note to be returned when
quering the day *before* the note creation.

The test case was tested against commit 55d5487, which was before this 
bug was resolved: the test failed (the note from 24th of Jan was returned when querying 
for notes created on the 23rd of jan)

Ran again against the current commit (post zk-org/zk#382 being resolved) and it
passes.

Also, in contributing.md, mickael wrote that it's good practice that we make a test fixture per issue with tesh tests. 
But there was already a tests/markdown-yaml-frontmatter.tesh file. 

Should I instead add this new test coverage to that test fixture?

### DIFF
--- a/tests/fixtures/issue-382/.zk/config.toml
+++ b/tests/fixtures/issue-382/.zk/config.toml
@@ -1,0 +1,197 @@
+# zk configuration file
+#
+# Uncomment the properties you want to customize.
+
+# NOTE SETTINGS
+#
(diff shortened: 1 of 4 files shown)


mcDevnagh (approved):
Beautiful

PR #387, opened 2024-01-25
Update documentation to restore GitHub Pages functionality

### LINKED ISSUE #386: GitHub Pages site does not have a homepage
The GitHub Pages site doesn't load an index.html, so the base URL throws a 404. I'll take a look at improving this behavior on my fork.

_Originally posted by @hugginsio in https://github.com/zk-org/zk/discussions/371#discussioncomment-8238313_

hugginsio:
This PR fixes the documentation site deployed to GitHub Pages, as pointed out in #386. You can view the site created from these changes at <https://docs.huggins.io/zk/>.

The easiest way to get the site working again was to move README.md to the `docs/` folder. This moves it out of the repository root, but it's still one of the standard README locations checked by GitHub - the README will still render at the bottom of the repository homepage. If we aren't okay with that move, I will explore making additional changes to the Pages deployment action so that we can leave the README where it is now.

To support this change, I have:

- Reverted to the default theme as Mickaël had in the unmerged `fix/gh-pages` branch.
- Added a Pages deployment action.
- Updated links in the README.
- Added the workflow description to the contribution guidelines.

Any and all feedback is appreciated. Thanks!

Closes #386.


mcDevnagh:
1. the README should still exist in the root of the repo, so that GitHub shows it
    - I'm fine with all pages, other than the README page, being prefixed with `/docs/` in the URL
2. Is there a way to do versioning?
     - I'm not convinced that `main` should be the only available documentation, which may change between releases, and thus give incorrect info to users of the latest version
     - it would be nice, if we had multiple versions of the docs available, based on the release tag, but if only the latest release is available, so be it

.github/workflows/gh-pages.yml
```diff
@@ -0,0 +1,43 @@
...
+        uses: actions/jekyll-build-pages@v1
+        with:
+          source: ./docs
+          destination: ./_site
+      - name: Upload artifact
+        uses: actions/upload-pages-artifact@v3
+  deploy:
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
```
mcDevnagh, line 37: This value gets read automatically from the steps below?

hugginsio: Yes, this is what powers the environment widget on the repository page. It's part of the default GitHub Pages Actions Workflow template.

https://docs.github.com/en/pages/getting-started-with-github-pages/using-custom-workflows-with-github-pages#deploying-github-pages-artifacts

hugginsio:
1. Should be able to do this with the default Jekyll static site generator. I'll look into this.
2. I'm not sure. I'll admit I haven't _really_ worked with GitHub Pages/Jekyll in a while. I'll look into this as well.

I'll add additional comments here and push updates to this branch once I've made progress on these two items.

hugginsio:
I was unable to find any nice Jekyll themes that handled versioned documentation well. We could roll our own solution with an existing theme and have every release as a subdirectory that you can navigate to in the sidebar, but that would make using any search functionality confusing.

[MkDocs][mkdocs] is a very popular SSG for project documentation these days, and it seems to support project documentation a little better than Jekyll does. The [Material for MkDocs][mkdocs-material] theme integrates well with a document versioning tool called [mike][mike]. At first glance, it looks like it would provide a nice workflow.

[mkdocs]: https://www.mkdocs.org
[mkdocs-material]: https://squidfunk.github.io/mkdocs-material/
[mike]: https://github.com/jimporter/mike

mcDevnagh:
The original solution will suffice, unless of course, you'd like to try to get versioning working. 
I would like the docs to only update on release though: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release

hugginsio:
If you're open to it, I'll happily take on a MkDocs refactor to get us versioned docs. However, that is outside the scope of this PR.

I have started work on fixing up the current GitHub Pages docs without moving the README around as we discussed. I will tag you here once it's ready for another review!

hugginsio:
Hello @mcDevnagh - I have pushed a few more changes. A deployment of the latest commit is live at <https://docs.huggins.io/zk/>. The README is back in the repository root, as requested.

mcDevnagh (approved):

mcDevnagh:
looks good. thank you kindly for your work here!

hugginsio:
Happy to help! Love this project and glad to be a (small) part of it.

tokisuno:
idk if this is still being worked on but https://docs.huggins.io/zk/ returns nothing. where should i be looking?

hugginsio:
The documentation is at https://zk-org.github.io/zk/.

I used my domain during development because I cannot make deployments in `zk-org`.

tokisuno:
> The documentation is at https://zk-org.github.io/zk/.
> 
> I used my domain during development because I cannot make deployments in `zk-org`.

oh nice. but is there any plan to implement some sort of search feature to the docs at all? or a table of contents to make scrolling around easier? some sort of way to source the /docs folder instead of having to do everything through the repo would be nice.

hugginsio:
@tokisuno - take a look at #388! There is some great discussion around redesigning the documentation there. Many of the potential directions there would result in better navigation and a search feature.

Issue #250, opened 2022-08-25
Broken file based url


skbolton:
I am getting [this](https://github.com/mickael-menu/zk/blob/ab1d8fd0bd7ab40648db14074a860eedf93f055c/internal/adapter/lsp/server.go#L602) diagnostic error from zk lsp in my notes for links to files. Here is an example of a link causing the issue.

`[sample](file:///home/skbolton/Documents/Notes/sample.pdf)`

If I use my terminal's functionality to go to the link under the cursor it opens the file just fine. If I change the format to:

`[sample](file://home/skbolton/Documents/Notes/sample.pdf)` zk no longer complains about it not finding the file but my terminal will no longer open the file.

From what I gather the `file:///` is the proper syntax for a file link on the current host and so it would be nice to have that working if possible so that I don't get an error in zk and my terminal properly visits the link.

> I am temporarily getting around this by doing the link as `[sample](file://localhost/home/skbolton/Documents/Notes/sample.pdf)` which is working for both systems (its just verbose).

mickael-menu:
Agreed, this kind of hrefs should be ignored by `zk`.

tjex:
Confirmed this is still happening, could potentially be looked at together with #194

PR #391, opened 2024-02-10
accept tripple dash file URIs as valid links

tjex:
PR to fix zk-org/zk#250. Links such as `[title](file:///file.go)` were returning an
lsp error "not found". The LSP does not actually check for the existence of
files via this format. Instead it was a symptom of the markdown link regex
checking, whereby the link was being understood as an internal markdown
document. Zk / lsp then tried to find the note `file:///file.example`, which it
if course couldn't, and so it was throwing the "not found" error.

This sollution copies the logic already present for embedded images, which is to
pick a defining syntactic element of the link (in this case, the third
occurrence of a '/' rune), and to skip that link with `continue`, effectively
ignoring it.

### DIFF
--- a/internal/adapter/lsp/document.go
+++ b/internal/adapter/lsp/document.go
@@ -7,12 +7,12 @@ import (
 	"strings"
 	"unicode/utf16"
 
+	"github.com/tliron/glsp"
+	protocol "github.com/tliron/glsp/protocol_3_16"
 	"github.com/zk-org/zk/internal/core"
 	"github.com/zk-org/zk/internal/util"
 	"github.com/zk-org/zk/internal/util/errors"
 	strutil "github.com/zk-org/zk/internal/util/strings"
-	"github.com/tliron/glsp"
-	protocol "github.com/tliron/glsp/protocol_3_16"
 )
 
 // documentStore holds opened documents.
@@ -228,23 +228,24 @@ func (d *document) DocumentLinks() ([]documentLink, error) {
 			})
 		}
 
-        // extract link paths from [title](path) patterns
-        for _, match := range 
-        markdownLinkRegex.FindAllStringSubmatchIndex(line, -1) {
-            // case: '!' ignores embedded image, e.g. ![title](href.png)
-            // case: '/' ignores tripple dash file URIs, e.g. file:///file.go
-			if match[0] > 0 && line[match[0]-1] == '!' || line[match[3]+8] == '/' {
+		// extract link paths from [title](path) patterns
+		for _, match := range markdownLinkRegex.FindAllStringSubmatchIndex(line, -1) {
+            // Ignore embedded images ![title](file.png) and tripple dash file 
+            // URIs [title](file:///file.go)
+			if (match[0] > 0 && line[match[0]-1] == '!') ||
+				(match[0] > 0 && line[match[3]+8] == '/') {
 				continue
 			}
 
 			href := line[match[4]:match[5]]
-			// Valid Markdown links are percent-encoded.
+
+			// Decode the href if it's percent-encoded
 			if decodedHref, err := url.PathUnescape(href); err == nil {
 				href = decodedHref
 			}
+
 			appendLink(href, match[0], match[1], false, false)
 		}
-
 
 		for _, match := range wikiLinkRegex.FindAllStringSubmatchIndex(line, -1) {
 			href := line[match[2]:match[3]]
--- a/internal/adapter/lsp/document.go
+++ b/internal/adapter/lsp/document.go
@@ -228,9 +228,12 @@ func (d *document) DocumentLinks() ([]documentLink, error) {
 			})
 		}
 
-		for _, match := range markdownLinkRegex.FindAllStringSubmatchIndex(line, -1) {
-			// Ignore embedded image, e.g. ![title](href.png)
-			if match[0] > 0 && line[match[0]-1] == '!' {
+        // extract link paths from [title](path) patterns
+        for _, match := range 
+        markdownLinkRegex.FindAllStringSubmatchIndex(line, -1) {
+            // case: '!' ignores embedded image, e.g. ![title](href.png)
+            // case: '/' ignores tripple dash file URIs, e.g. file:///file.go
+			if match[0] > 0 && line[match[0]-1] == '!' || line[match[3]+8] == '/' {
 				continue
 			}
 
@@ -241,6 +244,7 @@ func (d *document) DocumentLinks() ([]documentLink, error) {
 			}
 			appendLink(href, match[0], match[1], false, false)
 		}
+
 
 		for _, match := range wikiLinkRegex.FindAllStringSubmatchIndex(line, -1) {
 			href := line[match[2]:match[3]]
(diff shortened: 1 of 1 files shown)


tjex:
sorry for the formatting.. looks like the github cli is sending as formatted raw text rather than markdown compatible text... bit of a nish nish.

internal/adapter/lsp/document.go
```diff
@@ -228,17 +228,22 @@ func (d *document) DocumentLinks() ([]documentLink, error) {
...
 		}
 
+		// extract link paths from [title](path) patterns
 		for _, match := range markdownLinkRegex.FindAllStringSubmatchIndex(line, -1) {
-			// Ignore embedded image, e.g. ![title](href.png)
-			if match[0] > 0 && line[match[0]-1] == '!' {
+            // Ignore embedded images ![title](file.png) and tripple dash file 
+            // URIs [title](file:///file.go)
+			if (match[0] > 0 && line[match[0]-1] == '!') ||
+				(match[0] > 0 && line[match[3]+8] == '/') {
```
mcDevnagh, line 236: this is a very over-eager match
this matches the **/** here: file:/**/**/, but it would also match the following: ./dir1/file

Also, I don't think this is a good way to use the result of `FindAllStringSubmatchIndex`
- For the `markdownLinkRegex` (`\[([^\]]+?[^\\])\]\((.+?[^\\])\)`), it returns a slice of int slices
    - Each slice within the slice corresponds to a match
    - Each match will contain exactly 6 integers:
        - the starting index of the entire match (inclusive)
        - the ending index of the entire match (exclusive)
        - the starting index of the *bracket* capture group (inclusive)
        - the ending index of the *bracket* capture group (exclusive)
        - the starting index of the *parenthesis* capture group (inclusive)
        - the ending index of the *parenthesis* capture group (exclusive)
    - `match[3]` then corresponds to the index ending the _bracket_ capture group, which seems odd, since we're interested in the *parenthesis* capture group
        - `match[4]` is the one we should use 

Thus, I think we should match against the entire [scheme](https://en.wikipedia.org/wiki/File_URI_scheme)--`file://`--starting at `match[4]`. This is probably best done in a helper function, first checking that `match[5] - match[4]` is large enough for `file://` (8, since we need at least one character after the scheme to be a valid URI), then iterating from `match[4]` to `match[5]` matching each character along the way. If any character doesn't match, we return a `false` early.

Also, I know we're not explicitly supporting windows, but this will cause an obvious portability issue:
- POSIX has the third slash in `file:///` correspond to an actual file: the root file
- Windows seems to append an additional `/`, since `file:///c:` corresponds to the C drive, and `file:///` corresponds to nothing

I don't think we necessary need to support `file://` on Windows with this PR, but I am curious as to the result of this on Windows

internal/adapter/lsp/document.go
```diff
@@ -7,12 +7,12 @@ import (
...
 	"unicode/utf16"
 
+	"github.com/tliron/glsp"
+	protocol "github.com/tliron/glsp/protocol_3_16"
 	"github.com/zk-org/zk/internal/core"
 	"github.com/zk-org/zk/internal/util"
 	"github.com/zk-org/zk/internal/util/errors"
 	strutil "github.com/zk-org/zk/internal/util/strings"
-	"github.com/tliron/glsp"
-	protocol "github.com/tliron/glsp/protocol_3_16"
```
mcDevnagh, line 15: why the reordering? Did gofmt do this?

mcDevnagh (changes requested):

internal/adapter/lsp/document.go
tjex: yes. Same as the spaces / tabs you pointed out above... I went back to double check and ran `go fmt` again.

internal/adapter/lsp/document.go
tjex: Ahhh. I thought the ints followed the sequence in which they were found... as in 0 -> n from `[` -> `)`.
I'll have a proper look through all this in the coming days hopefully. But I'm on holiday for a month from Thursday!

Thanks so much for all the detail and thoroughness! 🙏

mcDevnagh: np. it took a lot of testing for me to figure out what exactly `FindAllStringSubmatchIndex` does. It's documentation leaves a lot to be desired

internal/adapter/lsp/document.go
mcDevnagh: sounds good. I'm fine with unrelated changes like this if it wasn't done properly before (i.e. `go fmt` wasn't run), but if its just a matter of preference, I'd rather keep the git blame clean. Looks like we're good to keep this

tjex: Nice, me too. I appreciate that level of attention / scrutiny 🔥

internal/adapter/lsp/document.go
tjex: @mcDevnagh 

Would this work, or is it in an ok direction?

```go
package main

import (
	"regexp"
)

const testCase = "[example](file:///foo.go)"

var markdownLinkRegex = regexp.MustCompile(`\[([^\]]+?[^\\])\]\((.+?[^\\])\)`)
var fileURIregex = regexp.MustCompile(`file:///`)

func main() {

	var result string
	for _, match := range markdownLinkRegex.FindAllStringSubmatchIndex(testCase, -1) {
		if match[5]-match[4] >= 8 {
			// get the file url portion, existing within the parentheses
			linkURL := testCase[match[4]:match[5]]
			// limit to first 8 characters as per 'file:///' scheme
			result = linkURL[:8]
		}
		// check if the first 8 characters of the URL match 'file:///'
		if fileURIregex.MatchString(result) {
			// if they match, we continue, meaning we DO NOT consider it an
			// internal note url, thus avoiding the lsp error that the note doesn't exist
			continue
		}
	}

}
```

tjex: I added this approach into document.go and tested. waiting for review.

jurica (commented):
Looks good to me.

tjex:
@mcDevnagh are you around to approve your reviewed changes (merging is blocked until you review)? Otherwise I'll dismiss and ask someone else to approve (just feels rude to dismiss until I try to ping you here as well).

mcDevnagh:
I'll take a look

tjex:
> Looks good to me.

@jurica , I think mcDevangh is a bit too snowed under to re-review. I've dismissed his review so that you can pass the review yourself. I'd like to move on to tackling the other bugs, which will be working in the same file as this PR.

jurica (approved):

jurica:
> > Looks good to me.
> 
> @jurica , I think mcDevangh is a bit too snowed under to re-review. I've dismissed his review so that you can pass the review as well, incase he needs more time. Just caus I'd like to move on to tackling the other bugs this long weekend, which will be working in the same file as this PR.

Sure if it's holding you back from doing more. I didn't want to merge it right away last week to give others the chance to also review...

mcDevnagh:
sorry for dropping the ball on this one! Thank you @jurica for picking up my slack. For what it's worth, code looks good :)

PR #395, opened 2024-03-15
Fix hyperlink from `README.md` to `docs/getting-started.md`

sumit-sahrawat:
This PR fixes an issue with the hyperlink to the getting-started doc in the GitHub pages generated site. The `href` in the HTML `div` is not mapped correctly, so we move the link out of the div into a separate line. Doesn't look as pretty but should work as expected.


mickael-menu (approved):
Thanks Sumit

Issue #194, opened 2022-04-23
LSP server is matching links inside fenced code blocks


mickael-menu:
_Originally posted by @daephx in https://github.com/mickael-menu/zk/issues/22#issuecomment-1107476645_

I don't know if this was mentioned but I didn't see any other relevant issues. It appears the LSP attempts to check wikilink syntax within code blocks. Which is probably a rare issue but still doesn't strike me as appropriate behavior.
I only noticed this due to the multi-line string syntax Lua uses.

![image](https://user-images.githubusercontent.com/36192863/164909034-2f384947-7670-4a5e-afd4-eeefb15cc0fd.png)

pidgeon777:
@mickael-menu are you using Vim/Neovim?

Could you please post your LSP config to enable zk inside it?

Best regards

mickael-menu:
My config is very basic. I'm using [`zk-nvim`](https://github.com/mickael-menu/zk-nvim) with:

```lua
require("zk").setup({
  picker = "telescope",

  lsp = {
    -- `config` is passed to `vim.lsp.start_client(config)`
    config = {
      cmd = {'zk', 'lsp', '--log', '/tmp/zk-lsp.log'},
      name = "zk",
    },

    -- automatically attach buffers in a zk notebook that match the given filetypes
    auto_attach = {
      enabled = true,
      filetypes = { "markdown" },
    },
  },
})
```

And `ftplugin` to add the mappings, as described here: https://github.com/mickael-menu/zk-nvim#example-mappings The sample `~/.config/nvim/ftplugin/markdown.lua` is mine.

tjex:
Confirmed this is still happening, could potentially be looked at together with #250

tjex:
Work for this is in progress by @Rahlir as mentioned [here in the same issue but over on zk-nvim](https://github.com/zk-org/zk-nvim/issues/161)

PR #399, opened 2024-04-02
fix(lsp): ignore diagnostic check within code blocks

Rahlir:
resolves: #194.
resolves: https://github.com/zk-org/zk-nvim/issues/161

I tried to be as thorough as possible, let me know if I missed any other markdown ways to denote code blocks. I am also including here two files I have used to test this fix:
[empty-note.md](https://github.com/zk-org/zk/files/14841478/empty-note.md)
[test-note.md](https://github.com/zk-org/zk/files/14841479/test-note.md)

### DIFF
--- a/internal/adapter/lsp/document.go
+++ b/internal/adapter/lsp/document.go
@@ -167,6 +167,11 @@ var fencedStartRegex = regexp.MustCompile(`^(` + "```" + `|~~~).*`)
 var fencedEndRegex = regexp.MustCompile(`^(` + "```" + `|~~~)\s*`)
 var indentedRegex = regexp.MustCompile(`^(\s{4}|\t).+`)
 
+var insideInline = false
+var insideFenced = false
+var insideIndented = false
+var currentCodeBlockStart = -1
+
 // LinkFromRoot returns a Link to this document from the root of the given
 // notebook.
 func (d *document) LinkFromRoot(nb *core.Notebook) (*documentLink, error) {
@@ -207,49 +212,53 @@ func linkWithinInlineCode(strBuffer string, linkStart, linkEnd int, insideInline
 	}
 }
 
+func ignoreLinesInCodeBlocks(lines []string, lineIndex int, line string) bool {
+	// Checks to ignore lines within code fences and indented code blocks
+	if insideFenced {
+		if fencedEndRegex.FindStringIndex(line) != nil &&
+			lines[currentCodeBlockStart][:3] == line[:3] {
+			// Fenced code block ends with this line
+			insideFenced = false
+			currentCodeBlockStart = -1
+		}
+		return true
+	} else if insideIndented {
+		if indentedRegex.FindStringIndex(line) == nil && len(line) > 0 {
+			// Indeted code block ends with this line
+			insideIndented = false
+			currentCodeBlockStart = -1
+		} else {
+			return true
+		}
+	} else {
+		// Check whether the current line is the start of a code fence or
+		// indented code block
+		if fencedStartRegex.FindStringIndex(line) != nil {
+			insideFenced = true
+			currentCodeBlockStart = lineIndex
+			return true
+		} else if indentedRegex.FindStringIndex(line) != nil &&
+			(lineIndex > 0 && len(lines[lineIndex-1]) == 0 || lineIndex == 0) {
+			insideIndented = true
+			currentCodeBlockStart = lineIndex
+			return true
+		}
+	}
+	return false
+
+}
+
 // DocumentLinks returns all the internal and external links found in the
 // document.
 func (d *document) DocumentLinks() ([]documentLink, error) {
 	links := []documentLink{}
 
-	insideInline := false
-	insideFenced := false
-	insideIndented := false
-	currentCodeBlockStart := -1
 	lines := d.GetLines()
 	for lineIndex, line := range lines {
 
-		// Checks to ignore lines within code fences and indented code blocks
-		if insideFenced {
-			if fencedEndRegex.FindStringIndex(line) != nil &&
-				lines[currentCodeBlockStart][:3] == line[:3] {
-				// Fenced code block ends with this line
-				insideFenced = false
-				currentCodeBlockStart = -1
-			}
-			continue
-		} else if insideIndented {
-			if indentedRegex.FindStringIndex(line) == nil && len(line) > 0 {
-				// Indeted code block ends with this line
-				insideIndented = false
-				currentCodeBlockStart = -1
-			} else {
-				continue
-			}
-		} else {
-			// Check whether the current line is the start of a code fence or
-			// indented code block
-			if fencedStartRegex.FindStringIndex(line) != nil {
-				insideFenced = true
-				currentCodeBlockStart = lineIndex
-				continue
-			} else if indentedRegex.FindStringIndex(line) != nil &&
-				(lineIndex > 0 && len(lines[lineIndex-1]) == 0 || lineIndex == 0) {
-				insideIndented = true
-				currentCodeBlockStart = lineIndex
-				continue
-			}
-		}
+        if ignoreLinesInCodeBlocks(lines, lineIndex, line) {
+            continue
+        }
 
 		appendLink := func(href string, start, end int, hasTitle bool, isWikiLink bool) {
 			if href == "" {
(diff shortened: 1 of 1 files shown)


tjex:
awesome! I'll have a look at this tomorrow (or latest day after)

tjex:
@Rahlir 
Can confirm that links within code blocks are ignored, as are links within inline back ticks; with one exception:

```text
16 |  `[[this is a link existing as the first string on a line]]`     ■ not found
```

These are getting missed for some reason. Will keep inspecting, but just wanted to chime in as I go.

Thanks again for this!

I do however think we can be clearer with the function and variable naming though. 
`linkWithinInlineCode` does in fact apply to code within inline back ticks _as well_ as code blocks. 

I would suggest simply `linkWithinBackticks` / `linkWithinCode` instead. Thoughts?

Rahlir:
Thanks for the feedback!

Hmm, weird that the first string is being missed, that's probably some oversight with 0 index, will fix that asap. Thanks for catching that.

Regarding the function names: I am not entirely sure what you mean. The function `linkWithinInlineCode` only checks whether the link is within backticks. The code block checks are separate checks done with regexes in the beginning of the for loop over lines. Or did I misunderstand what you meant?

tjex:
Found it :) It's because when the link is the first string in the line, the backtick == 0.

So `backtickId > 0 && backtickId < linkEnd` needs to be `backtickId >= 0 && backtickId < linkEnd`

Regarding the naming, that was my confusion! I thought the function was being used for code blocks as well. 
The comment perhaps (partially) threw me off (by my expertise had me already on a pretty diagonal course!)

Perhaps `// Recursive function to check whether a link is within inline code block.` could be changed to `// Recursive function to check whether a link is within inline code.`?

internal/adapter/lsp/document.go
```diff
@@ -194,11 +197,25 @@ func (d *document) DocumentLinkAt(pos protocol.Position) (*documentLink, error)
 	return nil, nil
 }
 
+// Recursive function to check whether a link is within inline code block.
+func linkWithinInlineCode(strBuffer string, linkStart, linkEnd int, insideInline bool) bool {
+	if backtickId := strings.Index(strBuffer, "`"); backtickId > 0 && backtickId < linkEnd {
```
tjex, line 202: `backtickId > 0` needs to be `backtickId >= 0` to account for links within backticks being the first instance of a string on the line.

internal/adapter/lsp/document.go
```diff
@@ -194,11 +197,25 @@ func (d *document) DocumentLinkAt(pos protocol.Position) (*documentLink, error)
 	return nil, nil
 }
 
+// Recursive function to check whether a link is within inline code block.
```
tjex, line 200: Tiny detail: as this function checks for _inline code_ exclusively, the comment should reflect that:

` ... a link is within inline code` / `... a link is within inline code / back ticks`

I think the term 'code block' has a general accepted meaning of a (multi-line) code block, and not inline code. It could therefore cause confusion down the track for someone.

tjex (changes requested):

internal/adapter/lsp/document.go
Rahlir: Huh, don't know what I was doing there. I think the equal sign got dropped accidentally when changing this function for the thousandth time :laughing:

internal/adapter/lsp/document.go
Rahlir: Yeah, you're totally right. `inline code block` is a bit contradictory...

tjex (approved):
great! Thanks for getting these changes back so quickly.

I'm going to give it a second pass with a fresh mind this morning, just to double check myself.

internal/adapter/lsp/document.go
```diff
@@ -259,10 +313,17 @@ func (d *document) DocumentLinks() ([]documentLink, error) {
...
 		for _, match := range wikiLinkRegex.FindAllStringSubmatchIndex(line, -1) {
+			// Ignore when inside backticks: `[[filename]]`
+			if linkWithinInlineCode(line, match[0], match[1], insideInline) {
+				continue
+			}
 			href := line[match[2]:match[3]]
 			hasTitle := match[4] != -1
 			appendLink(href, match[0], match[1], hasTitle, true)
 		}
+		if strings.Count(line, "`") % 2 == 1 {
```
tjex, line 324: @Rahlir I'm unsure what this `strings.Count()` check is doing? Is it perhaps left over from an initial solution? Because in addition to `linkWithinInlineCode()` it doesn't achieve anything or?

internal/adapter/lsp/document.go
```diff
@@ -229,10 +246,47 @@ func (d *document) DocumentLinks() ([]documentLink, error) {
 			})
 		}
 
+		if insideFenced {
```
tjex, line 249: Does it make sense to place this block above the declarations of `appendLink()`? Just to shave down some execution time? As otherwise, within a code block for example, `appendLink()` is being redeclared with data that will never be used, as that line will be fully skipped anyway, meaning that `appendLink()` won't even be called in its current declared state.

tjex (changes requested):

internal/adapter/lsp/document.go
```diff
@@ -229,10 +246,47 @@ func (d *document) DocumentLinks() ([]documentLink, error) {
...
+				currentCodeBlockStart = lineIndex
+				continue
+			} else if indentedRegex.FindStringIndex(line) != nil &&
+			(lineIndex > 0 && len(lines[lineIndex-1]) == 0 || lineIndex == 0) {
+				// Indented code block starts on this line
+				insideIndented = true
+				currentCodeBlockStart = lineIndex
+				continue
+			}
+		}
```
tjex, line 280: Does it make sense to place this block above the declarations of appendLink()? Just to shave down some execution time? As otherwise, within a code block for example, appendLink() is being redeclared with data that will never be used, as that line will be fully skipped anyway, meaning that appendLink() won't even be called in its current declared state.

internal/adapter/lsp/document.go
```diff
@@ -259,10 +313,17 @@ func (d *document) DocumentLinks() ([]documentLink, error) {
...
+			if linkWithinInlineCode(line, match[0], match[1], insideInline) {
+				continue
+			}
 			href := line[match[2]:match[3]]
 			hasTitle := match[4] != -1
 			appendLink(href, match[0], match[1], hasTitle, true)
 		}
+		if strings.Count(line, "`") % 2 == 1 {
+			insideInline = !insideInline
+		}
```
tjex, line 326: @Rahlir I'm unsure what this strings.Count() check is doing? Is it perhaps left over from an initial solution? Because in addition to linkWithinInlineCode() it doesn't achieve anything or?

tjex (changes requested):

internal/adapter/lsp/document.go
tjex, line 326: This is taking into account inline code over a line break right?
Is there a reason to assign the inverse here, instead of just `true`?

As I'm thinking if there are an odd number of backticks, the next line should _always_ begin with `insideInline = true` or?

I'm also wondering if there's ever a case where the line starts with `insideInline = true`, and there being an odd number of backticks. This would then make `insideInline = false` for the subsequent line.

internal/adapter/lsp/document.go
tjex, line 280: Does it make sense to place this block above the declaration of appendLink()? Just to shave down some execution time?

As otherwise, while on a line in a code block for example, appendLink() will be redeclared with data that will never be used. That line will be fully skipped anyway, meaning that the newly declared appendLink() won't even be called?

tjex:
@Rahlir I pushed the 'fixes' from my comments, so you can just pull to check them out if you like. If my comments are misguided, let me know and I'll revert!

Rahlir:
I think moving the check is fine and you are right that it might shave some execution time off.

Hardcoding `insideInline = true` is wrong though. The inversion is there because if you have three ticks on one line, the next line is still an inline comment, then if you have three backticks on next line, the third backtick ends the inline comment, it doesn't start it. That's why there is an inversion.

internal/adapter/lsp/document.go
tjex: addressed [in comment](https://github.com/zk-org/zk/pull/399#issuecomment-2039003713)

tjex:
ahhh yep, awesome, thanks for taking the time to explain to me 👌

Rahlir:
No problem :slightly_smiling_face:. All looks good to me now.

Issue #389, opened 2024-01-31
zk graph --format json is not handling note titles with inner double quotes


tjex:
### Check if applicable

- [X] I have searched the existing issues (**required**)
- [X] I'm willing to help fix the problem and contribute a pull request

### Describe the bug

I have four notes with inner double quotes in their titles, e.g. `This is a "note" title`.

Running `zk graph --json > notes.json` dumps json fine except for these notes. In their place, is a singular comma:

```json
    3 ▎   {filename:h5u1.md,filenameStem:h5u1,path:h.....<valid json continues>
E   2 ▎   ,     ■ Value expected
    1 ▎   {filename:jbig.md,filen......<valid json continues>
```

The weird thing is that there's only three instances of this, despite there being four notes with this quote pattern. Additionally, all four notes are not in the json file. In other words, three have been replaced with a singular column, and one has vanished altogether.

Additionally, it seems like notes with double quotes in their title create some strange behaviour. 
it is exclusively those notes that are printed to stdout, and after the printing of each note's body, my zsh environment variables are printed as well. Which is super weird, and would freak me out if I didn't know that this project is a legitimate project run by honest people haha.

I'm sure that will be fixed, once the strings get handled, but just wanted to put it out there.

### How to reproduce?

1. run `zk graph --json > notes.json` and check if it's already bug free
2. create a note `This is a "test" note`
3. run `zk index`
4. run `zk graph --json > notes2.json` 

Theoretically, there should be a singular line with a comma

### zk configuration

```toml
# zk configuration file
#
# Uncomment the properties you want to customize.

# NOTE SETTINGS
#
# Defines the default options used when generating new notes.
[note]
language = "en"
default-title = "untitled"
filename = "{{id}}"
extension = "md"
template = "default.md"

# Path globs ignored while indexing existing notes.
ignore = [
   "d/**",
]

# Configure random ID generation.
id-charset = "alphanum"
id-length = 4
id-case = "lower"

# EXTRA VARIABLES
#
# A dictionary of variables you can use for any custom values when generating
# new notes. They are accessible in templates with {{extra.<key>}}
[extra]

#key = "value"

# GROUP OVERRIDES
# Omitting `paths` is equivalent to providing a single path equal to the name of
# the group. This can be useful to quickly declare a group by the name of the
# directory it applies to.

[group.daily]
paths = ["d/2024"]

[group.daily.note]
filename = "{{format-date now}}"
extension = "md"
template = "daily.md"


[...]

Issue #48, opened 2021-06-17
Feature request: Plot graph using graphviz


pinpox:
It would be nice to be able to visualise the connections between notes as a graph. I have seen this functionality on other applications implementing that implement a Zettelkasten-style notebook and find it very helpful. Here is an [example of how that could look](https://github.com/MaxvdKolk/vimwikigraph#examples). 

I propose a command,  e.g. `zk graph` that, that just outputs the source for a graph as a graphviz graph, allowing to pipe it to a file or render it using `dot`.

Here is [another example](https://github.com/joashxu/zetteltools) of a very useful visualisation, haven't been able to run it on my `zk` notes folder though.

![image](https://user-images.githubusercontent.com/1719781/122354965-c63b2e80-cf51-11eb-99ae-e76545618737.png)

Maybe there already is a solution for this I'm not aware of, if so please let me know!

mickael-menu:
I had something similar in mind 👍 

I thought about having something like `zk graph --format <format> [filtering]`, using the same filtering options as `list` to get a subset of the graph.

Format could be `json`, `dot`, `svg` and more. And `ascii` format would be great, but that might stretch the terminal a bit too much!

An `--interactive` (xor with `--format`) mode serving an HTML version on the browser allowing to filter interactively the graph would be nice as well.

Initially I wanted to produce a JSON version, because it might be easier to pipe through other programs, including D3.js (which the project you shared uses). But I'm still trying to figure out the best JSON schema. There's a few spec out there but unfortunately nothing really standard:
* https://netflix.github.io/falcor/documentation/jsongraph.html
* http://jsongraphformat.info/
* JSON-LD too, but it might be a bit too complex for this use case

Making a prototype in DOT could be a first step. There's [even a Go package](https://github.com/emicklei/dot) to generate DOT graphs. I'm just wondering if DOT is really appropriate when you have so many nodes. But it could be useful for a small subset of your ZK.

pinpox:
Cool, let me know if you need help testing!

JSON would be very nice indeed, as it would allow to feed the output in something that renders mustache templates. I'm specifically thinking of [mustache-go](https://github.com/cbroglie/mustache) here, which I already use to render [static content](https://pinpox.github.io/nixos/), e.g. for [documentation](https://github.com/pinpox/nixos/blob/9abc10d193969ed019a09a4b2866d26739c4b5cf/manual/manual.nix#L40-L45).

This would also allow to just create a dot or svg mustache template probably without further code.

[...]

PR #400, opened 2024-04-03
fix quotes in titles breaking graph json output

tjex:
resolves: #389

note titles with `"` marks in them, eg `this is an "example" title`, are breaking the output of `zk graph --format json`.

The culprit is the template for the json format. I can't exactly track down where that is set, but it would seem deeper from `internal/core/notebook.go:336`.

In any case, this fix manually sanitizes the quote marks as they're being returned to the `Link` field, for output to json. Link output within documents (i.e, inserting wiki/markdown links in documents is not broken as that is handled separately in `internal/core/link_format.go`).

The caveat is that this fix breaks an old tesh test that looks at json output.

However, it would seem that the old tesh test is not taking into account note titles with `"` marks. 

Therefore, the failing test may be a sign of the test needing to be updated, rather than this fix being void?

In any case, this initial commit to the PR is effectively a draft. I'm looking for feedback about the above!

There is also [a discussion] in a previous issue which shows the development of the graph output, and also that [quotes in the titles are breaking its output](https://github.com/zk-org/zk/issues/48#issuecomment-863925549)

### DIFF
--- a/internal/core/note_format.go
+++ b/internal/core/note_format.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"regexp"
 	"time"
+    "strings"
 )
 
 // NoteFormatter formats notes to be printed on the screen.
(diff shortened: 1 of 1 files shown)


tjex:
@mickael-menu Would it be ok to get some high level input from you on this one? As you had [commented on the quotes breaking json before](https://github.com/zk-org/zk/issues/48#issuecomment-863925549). Also the usage of this template functionality has got me a bit stumped:

```go
func (n *Notebook) NewNoteFormatter(templateString string) (NoteFormatter, error) {
	templates, err := n.templateLoaderFactory(n.Config.Note.Lang)
	if err != nil {
		return nil, err
	}
	template, err := templates.LoadTemplate(templateString)
	if err != nil {
		return nil, err
	}

	linkFormatter, err := NewLinkFormatter(n.Config.Format.Markdown, templates)
	if err != nil {
		return nil, err
	}

	return newNoteFormatter(n.Path, template, linkFormatter, n.osEnv(), n.fs)
}
```

So I'm not that confident that I'm looking at this fix in the right way, and as the fix is changing the functionality that is also accessed by `zk list`, I'm wary of creating future bugs down the way.

tjex:
accidentally pushed here from `gh cli` instead of my own repo..

Issue #271, opened 2022-11-14
Notebook inside a hidden directory not working


ametis70:
I started using `zk` today on Linux with:
```
zk init "$HOME/.notes"
export ZK_NOTEBOOK_DIR="$HOME/.notes"
```
and then I created some notes with `new` and `edit --interactive`. After that, I found out that the `list` and `edit` commands couldn't find any note, even if the files were created correctly. I also tried with `index` without success.

After that, I moved the `$HOME/.notes` directory to `$HOME/Documents/Notes` and the commands now work correctly, so I believe there's a problem with the notebook being in a hidden directory

mickael-menu:
By design, [hidden files and directories are ignored from the `zk` index](https://github.com/mickael-menu/zk/blob/9d882451027b3e847cd6a6878d300e146387c6bb/internal/util/paths/walk.go#L24). Looks like this filter applies to the root notebook directory as well.

Until this is fixed, you can nest your notebook inside a hidden directory, e.g. `~/.notes/notebook`.

ametis70:
Alright, thanks!

tjex:
@ametis70 good to go :) Hope you're still using zk!

PR #402, opened 2024-04-07
allow notebook as hidden dir

tjex:
resolves: #271

The `index` function of `zk` calls a directory walk. That walk function was set to ignore all directories and files that had a period as a prefix. This applied then to the notebook root directory itself. This also effected commands such as `zk list`.

This fix is an initial approach to solving that by checking if the directory begins with a prefix _and_ the directory in question _does not_ equal the notebook root directory.

This equality check is done by getting the notebook directory from the `NotebookPath.Filename()` method, and checking it against the value of `filename` in `internal/util/paths/walk.go`.

Note: It was not possible to import the `core` package to `walk.go` as it created an import cycle. This is why it is instead accessed in `internal/core/note_index.go` line 153 and passed to the `paths.Walk()` call.

On that line (153), I'm unsure whether I've used the `NotebookPath` struct correctly as in other areas of the code base, _all_ fields of the struct are declared.

I have however tested this fix with success. That being a notebook with root dir `.test-vault`; `zk` indexes and lists correctly, meaning it ignores the `.zk` and `.git` directories as well as a `.test-file.md` note created via `zk`.

Please let me know if there is a better way to do this! 🙏

### DIFF
--- a/internal/core/note_index.go
+++ b/internal/core/note_index.go
@@ -150,7 +150,8 @@ func (t *indexTask) execute(callback func(change paths.DiffChange)) (NoteIndexin
 		return false, nil
 	}
 
-	source := paths.Walk(t.path, t.logger, shouldIgnorePath)
+    notebookPath := &NotebookPath{Path: t.path}
+	source := paths.Walk(t.path, t.logger, notebookPath.Filename(), shouldIgnorePath)
 
 	target, err := t.index.IndexedPaths()
 	if err != nil {
--- a/internal/core/note_index.go
+++ b/internal/core/note_index.go
@@ -150,7 +150,7 @@ func (t *indexTask) execute(callback func(change paths.DiffChange)) (NoteIndexin
 		return false, nil
 	}
 
-    notebookPath := &NotebookPath{Path: t.path}
+	notebookPath := &NotebookPath{Path: t.path}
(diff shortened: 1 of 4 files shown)


internal/core/note_index.go
```diff
@@ -150,7 +150,8 @@ func (t *indexTask) execute(callback func(change paths.DiffChange)) (NoteIndexin
 		return false, nil
 	}
 
-	source := paths.Walk(t.path, t.logger, shouldIgnorePath)
+    notebookPath := &NotebookPath{Path: t.path}
```
jurica, line 153: Please mind whitespaces. You're using spaces whereas in the file tabs are used.

jurica (commented):
@tjex I have one little nitpick.
Regarding your issue with the cyclic reference: IMHO it's perfectly fine to add that new argument to Walk.

internal/core/note_index.go
tjex: this is really weird. mcDevnagh picked up on the same issue from another PR. But I'm using gopls to format. Both from nvim via formatter.nvim and from the cli, gopls formats for me like this. 

not sure what to do about thath?

tjex: ups! forgot to run the formatter 😅
Sharp eyes!

jurica (approved):

PR #412, opened 2024-05-17
update Makefile with new images for linux go 1.21

tjex:
In preparation for v 0.14.1 release, so the correct images for linux builds are bundled with the release

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,13 +31,14 @@ dist-macos:
 	rm -f zk && make && zip -r "zk-${VERSION}-macos-`uname -m`.zip" zk
 
 # Produce a release bundle for Linux.
+# For documentation, see: https://github.com/zk-org/zk-xcompile
 dist-linux: dist-linux-amd64 dist-linux-arm64 dist-linux-i386
 dist-linux-amd64:
-	rm -f zk && docker run --rm -v "${PWD}":/usr/src/zk -w /usr/src/zk mickaelmenu/zk-xcompile:linux-amd64 /bin/bash -c 'make' && tar -zcvf "zk-${VERSION}-linux-amd64.tar.gz" zk
+	rm -f zk && docker run --rm -v "${PWD}":/usr/src/zk -w /usr/src/zk tjex/zk-xcompile:linux-amd64 /bin/bash -c 'make' && tar -zcvf "zk-${VERSION}-linux-amd64.tar.gz" zk
 dist-linux-arm64:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -31,13 +31,14 @@ dist-macos:
 	rm -f zk && make && zip -r "zk-${VERSION}-macos-`uname -m`.zip" zk
 
 # Produce a release bundle for Linux.
+# For documentation, see: https://github.com/zk-org/zk-xcompile
 dist-linux: dist-linux-amd64 dist-linux-arm64 dist-linux-i386
 dist-linux-amd64:
-	rm -f zk && docker run --rm -v "${PWD}":/usr/src/zk -w /usr/src/zk mickaelmenu/zk-xcompile:linux-amd64 /bin/bash -c 'make' && tar -zcvf "zk-${VERSION}-linux-amd64.tar.gz" zk
+	rm -f zk && docker run --rm -v "${PWD}":/usr/src/zk -w /usr/src/zk tjex/zk-xcompile:linux-amd64 /bin/bash -c 'make' && tar -zcvf "zk-${VERSION}-linux-amd64.tar.gz" zk
```
mcDevnagh, line 37: show this be the zk org?

Makefile
```diff
@@ -58,3 +79,8 @@ define go
 	$(ENV_PREFIX) go $(1) -tags "fts5" -ldflags "-X=main.Version=$(VERSION) -X=main.Build=$(BUILD)" $(2)
 endef
 
+# Alpine (musl) requires statically linked libs. This should be compatible for
```
tjex, line 82: not sure if this is the cleanest way to simply add `-extldflags=-static` to the build case for alpine? But couldn't come up with anything else.

Makefile
```diff
@@ -58,3 +79,8 @@ define go
...
+# Void linux and other musl based distros aswell.
+define alpine
+	$(ENV_PREFIX) go $(1) -tags "fts5" -ldflags "-extldflags=-static -X=main.Version=$(VERSION) -X=main.Build=$(BUILD)" $(2)
+endef
```
tjex, line 86: not sure if this is the cleanest way to simply add -extldflags=-static to the build case for alpine? But couldn't come up with anything else.

tjex:
It's necessary to confirm and merge https://github.com/zk-org/zk-xcompile/pull/5 first, as that defines the naming scheme for the alpine binaries / images, which will need to be reflected in the zk Makefile.

tjex:
I'm not meaning to be too 'executive' about merging this myself. But I've got a trip to Japan next week and really want to get this release out before then. I also figured it was part of the reasoning behind suggesting I become an owner of the org. Let me know if anything is out of line there 👍

mickael-menu:
That's fine by me, especially for maintenance work. Enjoy your trip in Japan!

Issue #411, opened 2024-05-17
`musl` systems not supported


notramo:
### Check if applicable

- [X] I have searched the existing issues (**required**)
- [ ] I'm willing to help fix the problem and contribute a pull request

### Describe the bug

`musl` is an alternative to `glibc`, and is used by multiple Linux distros e.g. Alpine Linux and Void Linux.

The official builds link to `glibc`, which is incompatible with `musl`-based systems. I recommend statically linking `musl` as it's compatible with both `glibc` and `musl`.

### How to reproduce?

1. Install Alpine Linux.
2. Download the built release of this software.

### zk configuration

```toml
does not matter
```

### Environment

```bash
any `musl` based Linux distro
```

tjex:
We cross compile with Docker images. If you change your mind about wanting to contribute, we'd welcome a PR to [zk-xcompile](https://github.com/zk-org/zk-xcompile) (from those current images, you can easily infer what would need to be in an Alpine based image, also better than me personally as I have no experience with `musl`) with a golang 1.21 image based from [Golang's Alpine images](https://hub.docker.com/_/golang/tags?page=&page_size=&ordering=&name=alpine).

For the next release of `zk` we could well have migrated away from needing Docker images for cross compilation, and will be able to do it natively. As the next release is potentially a few hours away `0.14.1`, I'm resistant to setting up an Alpine Docker image for this (potentially) singular minor release.

Alternatively, building `zk` from source is very user friendly (thanks to Go). So you can always go via that route.

Clone `zk`, `cd zk` and `make install`.

tjex:
Was curious, and had a bit of a go: https://github.com/tjex/zk-xcompile/tree/alpine

But attempting to compile results in many complaints about sqlite bindings (which is the dep that requires CGO).
Happy to work with you or someone else on this or receive a direct PR.

notramo:
Upgrading `go-sqlite3` to 1.14.19 should fix the issue:
https://github.com/mattn/go-sqlite3/releases/tag/v1.14.19
(related [issue](https://github.com/mattn/go-sqlite3/issues/1164) and [PR](https://github.com/mattn/go-sqlite3/pull/1177))

tjex:
Ah great, thanks for working with me to get the necessary info.

Bumping sqlite version built successfully with golang 1.21 and latest alpine build.

`file` returns this info:

[...]

PR #414, opened 2024-05-19
raise sqlite version for musl build compatibility

tjex:
resolves: #411

Our current version (v1.14.16) of `go-sqlite` breaks cross compilation to `musl` based distros such as Alpine.

We don't currently build for Alpine, but we can with this version bump.

I have an untested successful Alpine build in progress, just waiting on a report back whether it runs (see discussion in issue above).

Using the image in the [alpine branch of zk-xcompile](https://github.com/tjex/zk-xcompile/tree/alpine)


mcDevnagh (approved):

mcDevnagh:
Once you confirm the Alpine build, this can be merged

PR #428, opened 2024-07-07
Fix mix-up in error message when aborting

WhyNotHugo:
If I abort editing a file, `zk` shows 'failed to launch editor'. This message is not correct, since I explicitly aborted the operation after the editor successfully launched.

When using Vim, this is triggered by using `:cq`, which exits non-zero. With a GUI editor, this is usually achieved by pressing the Cancel/Abort button.

Update the error handling to show the correct error: if the editor executed and exited abnormally, indicate that the operation was aborted. Otherwise, show the previous message.

### DIFF
--- a/internal/adapter/editor/editor.go
+++ b/internal/adapter/editor/editor.go
@@ -3,6 +3,7 @@ package editor
 import (
 	"fmt"
 	"os"
+	"os/exec"
 	"strings"
 
 	"github.com/kballard/go-shellquote"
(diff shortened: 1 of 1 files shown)


tjex:
Can you elaborate on the steps?
I do not get this error if I:

1. `zk edit <note.md>`
2. quit the editor

tjex:
`:cq` exits with error code 1. So this is expected behaviour:

```text
							*:cq* *:cquit*
:cq[uit][!]
:{N}cq[uit][!]
:cq[uit][!] {N}		Quit Vim with error code {N}.  {N} defaults to one.
			Useful when Vim is called from another program:
			e.g., a compiler will not compile the same file again,
			`git commit` will abort the committing process, `fc`
			(built-in for shells like bash and zsh) will not
			execute the command, etc.
			{N} can also be zero, in which case Vim exits
			normally.
			WARNING: All changes in files are lost.  It works like
			":qall!" |:qall|, except that Nvim exits non-zero or
			[count].
```

WhyNotHugo:
Right, `:cq` aborts (exist non-zero).

So the error message `failed to launch editor` is incorrect, the editor launched fine and the operation was manually aborted.

What's wrong with the patch?

tjex:
ahhhh yes. Sorry, I overlooked that. 
Thanks 🙂

tjex (approved):
Nice optimisation. Thanks

PR #429, opened 2024-07-09
fix: magnet link

billymosis:
- Added check for magnet links (magnet:?xt) to prevent not found false alarms in LSP.

Context:
image

### DIFF
--- a/internal/adapter/lsp/document.go
+++ b/internal/adapter/lsp/document.go
@@ -206,6 +206,7 @@ var fileURIregex = regexp.MustCompile(`file:///`)
 var fencedStartRegex = regexp.MustCompile(`^(` + "```" + `|~~~).*`)
 var fencedEndRegex = regexp.MustCompile(`^(` + "```" + `|~~~)\s*`)
 var indentedRegex = regexp.MustCompile(`^(\s{4}|\t).+`)
+var magnetRegex = regexp.MustCompile(`magnet:\?`)
 
 var insideInline = false
 var insideFenced = false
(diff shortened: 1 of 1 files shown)


tjex:
Thanks for the pr :)

The approach so far for appeasing the LSP is to ignore urls that are not links to notes. https://github.com/zk-org/zk/pull/391 is an example of this.

Here in `document.go` 
https://github.com/zk-org/zk/blob/739d2d4fc03512d439fd20cad1eb4caabcbe6c67/internal/adapter/lsp/document.go#L295

I think it's still a better approach than adding a magnet link scheme to `IsURL` because 

a) it reduces load on the lsp and 
b) it doesn't dilute the meaning of `IsURL`. 

While a magnet link is in essence a url, for our purposes, it may be cleaner to just ignore it rather than validate it? Happy for any counter thoughts.

billymosis:
Thanks for the feedback!

I missed the file and line you mentioned in document.go. I see your point about ignoring URLs that are not links to notes, as shown in PR #391.

Just to provide more context that I use obsidian  and zk . And on the obsidian there was no linting error. And I could just click the magnet link easily. If I were using exclamation on the front and using back tick it will clutter my obsidian view.

What if I add something like this on existing `file:///` check

```go
			// ignore tripple dash file URIs [title](file:///foo.go)
			if match[5]-match[4] >= 8 {
				linkURL := line[match[4]:match[5]]
				fileURIresult := linkURL[:8]
				if fileURIregex.MatchString(fileURIresult) {
					continue
				}
                        // Ignore magnet
				if regexp.MustCompile(`magnet:?`).MatchString(fileURIresult) {
					continue
				}
			}

```

tjex:
That's quite convenient that it's also 8 chars.

May as well add it in the same `if` statement, to reduce branching and keep them together.
And it would be nicer to declare the regex for the magnet link as a variable beforehand (up in the file where the others are also declared), to maintain readability and reusability.

So: 

```go

magnetRegex := regexp.MustCompile(`magnet:?`)

			// ignore tripple dash file URIs [title](file:///foo.go) and magnet links
			if match[5]-match[4] >= 8 {
				linkURL := line[match[4]:match[5]]
				fileURIresult := linkURL[:8]
				if fileURIregex.MatchString(fileURIresult) || if magnetRegex.MatchString(fileURIresult) {
					continue
				}
			}

```

Double check you don't need to escape the `?` char in the regex statement!

billymosis:
I've updated the commit. Let me know if adjustments are needed.

tjex (approved):
thanks :)

tjex (approved):

Issue #419, opened 2024-05-26
v0.14.1 homebrew release action failed


tjex:
### Check if applicable

- [X] I have searched the existing issues (**required**)
- [ ] I'm willing to help fix the problem and contribute a pull request

### Describe the bug

Lodging this for now, to look at tomorrow. It's late in the evening 🌙

Action: https://github.com/zk-org/zk/actions/runs/9242929309/job/25426386503

### How to reproduce?

na

### zk configuration

```toml
na
```

mickael-menu:
I fixed the issue by [creating a new token](https://github.com/settings/tokens) and modifying the repository secret `HOMEBREW_GITHUB_TOKEN` in Settings > Secrets and variables > Actions.

However the 0.14.1 [was pushed manually on Homebrew by a contributor](https://github.com/Homebrew/homebrew-core/pull/172886) so the GitHub action still failed. That's fine as it's published. 👌

tjex:
@mickael-menu Awesome. Was also looking into this and getting a bit confused about where that token needed to be set.
However, I don't see any token or variable in the orgs settings.

Or is it that we should each generate a token for this github action?

The link takes me to my personal settings/tokens, rather than those for the orgs, which also confused the hell out of me from GitHub documentation.

https://github.com/settings/tokens vs https://github.com/organizations/zk-org/settings/secrets/actions

<img width="989" alt="Screenshot 2024-05-27 at 18 13 40" src="https://github.com/zk-org/zk/assets/83029642/56800a92-c372-4c7b-9715-a45adbe3f9e3">

mickael-menu:
The secrets are per project, so it's in the Settings of the zk repo itself. But I guess it could work in the org secrets as well. 
I'm not sure we can provide a token from the organization itself, I used my personal account to generate the token. I didn't give any access to my personal info or private repo to the token, so that should be fine.

tjex:
oh man. derp. of course 😅

Thanks :)

PR #430, opened 2024-07-09
gh-pages build on push, not release

tjex:
The gh pages fix from https://github.com/zk-org/zk/pull/387 setup deployment / build to occur on release publishing.

The documentation / readme is therefore stuck until the next zk semver release.

I think it's better to have the page rebuild on push / merge, as documentation updates are seemingly frequent.

@hugginsio what do you think?

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -3,8 +3,6 @@ name: Deploy to GitHub Pages
 on:
   push:
     branches: [ main ]
-  pull_request:
-    branches: [ main ]
 
 permissions:
   contents: read
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -1,9 +1,10 @@
 name: Deploy to GitHub Pages
 
 on:
-  release:
-    types:
-      - "published"
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
 
 permissions:
   contents: read


hugginsio:
Hi @tjex - nice catch!

My reasoning behind only rebuilding on release was so that documentation doesn't get out of sync with the application features. I believe this could be extended to also include pattern matching so that README updates trigger a rebuild.

There might be a place for that down the road, but while documentation gets updated (and maybe eventually overhauled) rebuilding the site on every push is probably better.

hugginsio:
> Looks like it's not as simple as changing to on-push. https://github.com/zk-org/zk/actions/runs/9860060569/job/27225298529?pr=430

That is an unusual error... especially considering we have the `id-token` set properly, same as the examples?

https://github.com/zk-org/zk/blob/739d2d4fc03512d439fd20cad1eb4caabcbe6c67/.github/workflows/gh-pages.yml#L8-L11

Did anyone make changes to the Actions settings at the organization level since we last deployed to Pages?

tjex:
Not at the organisation level. We actually never had any secrets / tokens at the org level.
The only time I remember talking about secrets was here: https://github.com/zk-org/zk/issues/419#issuecomment-2132978341

But it was to do with homebrew.

Thing is, apart from the homebrew secret. We don't have any other tokens...?
There's also no environment secrets set for the gh-pages environment settings in `settings -> environments`

So I'm actually a bit confused how pages even deployed in the first place 🫤

.github/workflows/gh-pages.yml
```diff
@@ -1,9 +1,10 @@
 name: Deploy to GitHub Pages
 
 on:
-  release:
-    types:
-      - "published"
+  push:
+    branches: [ main ]
+  pull_request:
```
mcDevnagh, line 6: we shouldn't deploy on PR. Only if `main` is updated

mcDevnagh: this is causing the workflow to fail. Random PRs should **not** have permissions to deploy to our github pages. Anyone who can open a PR could then do anything. The workflow should work on main

mcDevnagh (changes requested):
we shouldn't deploy on PR. Only if `main` is updated

.github/workflows/gh-pages.yml
tjex: ahhh right. I thought otherwise PRs for docs wouldn't go through... But of course, this just gets run as soon as the PR is lodged. GREAT catch

mcDevnagh (approved):

mcDevnagh:
https://github.com/zk-org/zk/actions/runs/9895195879
success

tjex:
Great work all! That was a snappy turn around 💫

PR #431, opened 2024-07-09
Expand tilde in config note.template

WhyNotHugo:
This allows specifying paths relative to $HOME in the configuration file (this is particularly useful for the global config).

    [note]
    template = "~/.config/zk/default.md"

### DIFF
--- a/internal/core/config.go
+++ b/internal/core/config.go
@@ -8,6 +8,7 @@ import (
 	toml "github.com/pelletier/go-toml"
 	"github.com/zk-org/zk/internal/util/errors"
 	"github.com/zk-org/zk/internal/util/opt"
+	"github.com/zk-org/zk/internal/util/paths"
 )
 
 // Config holds the user configuration.
@@ -312,7 +313,11 @@ func ParseConfig(content []byte, path string, parentConfig Config, isGlobal bool
 		config.Note.Extension = note.Extension
 	}
 	if note.Template != "" {
-		config.Note.BodyTemplatePath = opt.NewNotEmptyString(note.Template)
+		expanded, err := paths.ExpandTilde(note.Template)
+		if err != nil {
+			return config, wrap(err)
+		}
+		config.Note.BodyTemplatePath = opt.NewNotEmptyString(expanded)
 	}
 	if note.IDLength != 0 {
 		config.Note.IDOptions.Length = note.IDLength
--- a/internal/core/notebook_store.go
+++ b/internal/core/notebook_store.go
@@ -201,7 +201,7 @@ const defaultConfig = `# zk configuration file
 #extension = "md"
 
 # Template used to generate a note's content.
-# If not an absolute path, it is relative to .zk/templates/
+# If not an absolute path or "~/path", it's relative to .zk/templates/
 template = "default.md"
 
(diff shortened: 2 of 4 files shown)


tjex:
Thanks for the pr. I also much prefer to support `~` for dir decleration, but.. if it's to be implemented, then it then it needs to be baked in globally for the config. Not on a case by case basis. Otherwise users will get the idea that `~` is supported, only to have their confs break elsewhere, and then report a bug.

As there are other path settings in the conf, and they will break if provided a tilde.

This works:
```toml
[alias]
conf = '$EDITOR "/Users/tjex/.local/src/zk-org/workbench/test-vault/.zk/config.toml"'
```

This breaks:

```toml
[alias]
conf = '$EDITOR "~/.local/src/zk-org/workbench/test-vault/.zk/config.toml"'
```

WhyNotHugo:
`journal = "zk new --no-input ~/zk/journal/"` works fine, because this is passed to `sh -c`, which does the tilde expansion itself.

If you put it in quotes it doesn't expand, but that's kind of the point of putting it in quotes, no?

tjex:
Yes, fair enough 👍

Additionally though, I think the expandTilde function should be made globally available from [paths.go](https://github.com/zk-org/zk/blob/main/internal/util/paths/paths.go). Probably also the right time to beef build it as `ResolveHomePath(path String)` (or an alternate name in that direction)? May as well add a check for `$HOME`?

Will no doubt come in handy elsewhere.
Would you be up for it?

The default config should also be updated to reflect these changes: https://github.com/zk-org/zk/blob/6134873a5a5328484307403a1ed75625c1fff1cb/internal/core/notebook_store.go#L181

Meaning then the tesh test case should be updated (I can do that and the above if you like): https://github.com/zk-org/zk/blob/6134873a5a5328484307403a1ed75625c1fff1cb/tests/cmd-init-defaults.tesh

WhyNotHugo:
I moved `ExpandTilde` into the paths module :+1:

I'm not sure that I understand the rest of what you mean. In the following case, the quotes are there so that if `$ZK_NOTEBOOK_DIR` has spaces, the shell will treat the whole path as a single argument.

```
# Edit this configuration file.
#conf = '$EDITOR "$ZK_NOTEBOOK_DIR/.zk/config.toml"'
```

Removing the quotes would break situations where `$ZK_NOTEBOOK_DIR` contains spaces.

tjex:
> Removing the quotes would break situations where `$ZK_NOTEBOOK_DIR` contains spaces.

Totally true. Another good catch 🤦‍♂️ Sorry, juggling a bit too much this week.

Thanks for moving the function. I'll check it out and then push changes for the comments in the config, so users know they have more options than absolute path or relative path to .zk/conf for their path declaration.

i.e,

https://github.com/zk-org/zk/blob/6134873a5a5328484307403a1ed75625c1fff1cb/internal/core/notebook_store.go#L203-L204

And update the [cmd-init-defaults.tesh](https://github.com/zk-org/zk/blob/6134873a5a5328484307403a1ed75625c1fff1cb/tests/cmd-init-defaults.tesh) file accordingly.

tjex (approved):
Thanks again :)

Issue #413, opened 2024-05-18
`zk init` generates ambigous LSP section in config


notramo:
### Check if applicable

- [X] I have searched the existing issues (**required**)
- [ ] I'm willing to help fix the problem and contribute a pull request

### Describe the bug

The LSP section in the `zk init` generated config file is ambigous.
I don't know what are the possible values for the `lsp.completion` section.

Are the possible values the empty strings? (so it's enough to delete the comment marks at the line beginning?)
```toml
[lsp.completion]
# Customize the completion pop-up of your LSP client.

# Show the note title in the completion pop-up, or fallback on its path if empty.
#note-label = ""
# Filter out the completion pop-up using the note title or its path.
#note-filter-text = " "
# Show the note filename without extension as detail.
#note-detail = ""
```

### How to reproduce?

Run `zk init` and see `.zk/config.toml`

### zk configuration

```toml
Vanilla config generated by `zk init`
```

### Environment

```bash
any
```

tjex:
lsp config is documented [here in the repo](https://github.com/zk-org/zk/blob/main/docs/config-lsp.md).

I agree it would be good to populate these with sane defaults, and leave them commented out.

Starting point would [be here](https://github.com/zk-org/zk/blob/bf5cad60dcd0d315eba7b6d43075ff7f9332d086/internal/core/config.go#L56)

tjex:
It's because the braces aren't escaped. Fix incoming.

```text
# Show the note title in the completion pop-up, or fallback on its path if empty.
#note-label = "{{title-or-path}}"
```

should be 

```text
# Show the note title in the completion pop-up, or fallback on its path if empty.
#note-label = "\{{title-or-path}}"
```

As in other parts of the default config:

```text
#filename = "\{{id}}"
```

PR #432, opened 2024-07-11
fix: lsp completion config defaults not rendering

tjex:
Resolves zk-org/zk#413

Braces for config options needed to be escaped.


tjex:
pfff. really roasted the commits. closing to pr again.

PR #442, opened 2024-08-07
Refactor tilde expansion

tjex:
We added tilde expansion for note.template so users can share template files between notebooks in https://github.com/zk-org/zk/pull/431.

We however overlooked that tilde / home dir expansion was already being implemented in isolation for the global notebook directory in container.go.

I've refactored that code into the function @WhyNotHugo created, and also made a few optimisations to it (and renamed it as it now expands more than just tildes).

Otherwise, am I overlooking anything, have I overdone / underdone anything, or is it looking alright?


tjex:
I'm really stumped by the failing tesh test.. It's not failing on my end. I've double checked that there have been no changes between my local branch and this remote branch...

tjex:
`make tesh-update` passes CI.
`make tesh` does not.

hugo:
Might help you to tag the right Hugo if you want constructive feedback 😄 (Not a complaint/criticism, just you may actually want them to see this.)

tjex:
woops! thanks 😅 haha

internal/core/config.go
```diff
@@ -313,7 +313,7 @@ func ParseConfig(content []byte, path string, parentConfig Config, isGlobal bool
 		config.Note.Extension = note.Extension
 	}
```
tjex, line 314: @WhyNotHugo 
It occurred to me that if there is an error, then we will cease filling the rest of the config entirely.
This didn't occur to me when reviewing #431. I guess it's acceptable, as `user.Current()` should always be resolvable, otherwise something is seriously wrong?

Otherwise, we could just wrap the error and continue? What would be the drawback of that approach?

internal/core/config.go
```diff
@@ -313,7 +313,7 @@ func ParseConfig(content []byte, path string, parentConfig Config, isGlobal bool
...
 	if note.Template != "" {
-		expanded, err := paths.ExpandTilde(note.Template)
+		expanded, err := paths.ExpandHomeDir(note.Template)
 		if err != nil {
 			return config, wrap(err)
 		}
```
tjex, line 319: @WhyNotHugo
It occurred to me that if there is an error, then we will cease filling the rest of the config entirely.
This didn't occur to me when reviewing https://github.com/zk-org/zk/pull/431. I guess it's acceptable, as user.Current() should always be resolvable, otherwise something is seriously wrong?

Otherwise, we could just wrap the error and continue? What would be the drawback of that approach?

tjex:
Suspicions were correct. GitHub CI doesn't have `$HOME` set. So checking directly for `os.UserHomeDir()` was breaking GitHub CI, while it was passing fine locally.

Solution is then exactly as @WhyNotHugo did it 😅. First check for `os.CurrentUser()`, which seemingly works for GitHub CI and then access the `HomeDir` field of the returned struct. It does not error if there is none, so the function continues executing fine and the path string passed to `ExpandHomeDir()` is returned unchanged.

I guess there was no issue using `os.UserHomeDir()` [here in `container.go`](https://github.com/zk-org/zk/blob/4a51e398ae4572b2a1338fd9017ceeff30e5ffda/internal/cli/container.go#L77-L87), because there were no `tesh` cases for it. That threw me off for a while...

P.S The force push was just hard resetting to the previous commit in a lazy way... I know it's bad...

internal/util/paths/paths.go
```diff
@@ -79,17 +78,24 @@ func WriteString(path string, content string) error {
...
+func ExpandHomeDir(path string) (string, error) {
+
+	// Expand in case there are environment variables on the path
+	path = os.ExpandEnv(path)
 	usr, err := user.Current()
 	if err != nil {
-		return "", fmt.Errorf("failed to determine current user")
+		return "", err
 	}
 	home := usr.HomeDir
```
mcDevnagh, line 92: > HomeDir is the path to the user's home directory (_if they have one_).

https://pkg.go.dev/os/user#User emphasis mine

If a user doesn't have a home, but we need one (`~` is specified), we should create an error. We shouldn't be using an empty string as `home`

internal/util/paths/paths.go
```diff
@@ -79,17 +78,24 @@ func WriteString(path string, content string) error {
...
-func ExpandTilde(path string) (string, error) {
+// Expands home directory to absolute path.
+// If no expansion is required or possible, the string is returned unaltered.
+// If there is an error, an empty string is returned.
+func ExpandHomeDir(path string) (string, error) {
+
+	// Expand in case there are environment variables on the path
+	path = os.ExpandEnv(path)
 	usr, err := user.Current()
 	if err != nil {
```
mcDevnagh, line 89: we only need the `usr` if `~` is present in the string. It would be perfectly valid to have an absolute path while running without a user (I don't know the use case of this, but it would be good future proofing)

internal/util/paths/paths.go
```diff
@@ -79,17 +78,24 @@ func WriteString(path string, content string) error {
 	return err
 }
 
-// Expands leading tilde.
-func ExpandTilde(path string) (string, error) {
+// Expands home directory to absolute path.
+// If no expansion is required or possible, the string is returned unaltered.
+// If there is an error, an empty string is returned.
+func ExpandHomeDir(path string) (string, error) {
```
mcDevnagh, line 84: should this just be called `Expand`? You're expanding the `~` and any environment variables

internal/util/paths/paths.go
mcDevnagh: if `~` is present _and_ `err != nil`, then we `return "", err`

internal/util/paths/paths.go
jurica: I second that.

internal/util/paths/paths.go
jurica: I would suggest `ExpandTileAndEnv`, imho `Expand` is not specific enough.

tjex: Or `ExpandPath`? The `..Env` part could also mean an "environment" in some other context than a shell environment variable? It's reasonable to understand that `ExpandPath` would include tilde expansion?

```go
// Expands environment variables and `~`, returning an absolute path.
func ExpandPath(path string) (string, error) {}
```

internal/util/paths/paths.go
tjex: I'm confused why we should return an empty string if `~` is present and the current user is resolvable? 

In this case, the user would just using their home directory for a notebook / some other path. Which is valid.

tjex: @mcDevnagh ?

internal/core/config.go
tjex: @WhyNotHugo ?

internal/util/paths/paths.go
```diff
@@ -79,17 +79,30 @@ func WriteString(path string, content string) error {
...
+		if err != nil {
+			return "", err
+		}
+		home := usr.HomeDir
+		if home == "" {
+			return "", errors.New("Could not find user's home directory.")
+		}
+
+		// construct as abs path
+		if path == "~" {
```
WhyNotHugo, line 98: This condition is always false, because this is inside `if strings.HasPrefix(path, "~/") {`.

You probably want: `if strings.HasPrefix(path, "~") {`. above.

internal/util/paths/paths.go
```diff
@@ -79,17 +79,30 @@ func WriteString(path string, content string) error {
...
+			path = home
+		} else if strings.HasPrefix(path, "~/") {
+			path = filepath.Join(home, path[2:])
+		}
 	}
+
```
WhyNotHugo, line 104: Not sure if we want to support `~user/some/path` too, but this currently simply remains uncapped.

internal/util/paths/paths.go
tjex: ooh yes 👌

internal/util/paths/paths.go
tjex: I don't think it's necessary to support this. I've never actually seen that syntax.

PR #443, opened 2024-08-07
simple change for gh CI build check. This should pass

tjex:
build action in #442 is failing in github CI, but is not failing locally for me. Using this PR, which is at latest state of main to check if github CI is playing up somehow.


tjex:
it passed...

PR #450, opened 2024-08-22
Add feature: zk list --tagless

njnygaard:
Add support for `--tagless` to filter for notes that do not have tags per this discussion: https://github.com/zk-org/zk/discussions/449


tjex:
Hey. Thanks for following through on this :) 

I've just had a late night scan of it and it all looks straight forward and is making sense. I'll double check it with a fresher mind tomorrow ideally. 

And A+ for adding the tesh cases 👏

tjex (approved):
All looks good. I tweaked the flag description so that it reads the same as other descriptions, and also added the feature to some other docs. Thanks again! A great feature contribution!

PR #452, opened 2024-09-09
Create create-edit-list-projects-example.md

Zethtren:
Didn't see an example for this workflow but, I coded it up for personal use and figured I'd share it back in case anyone else wants it. 

Feel free to ignore this if you don't like it.


tjex:
Thanks for writing this up. It's great to see how people use zk. But I think such a document is better for the ["share" discussions section](https://github.com/zk-org/zk/discussions/categories/share).

Otherwise the docs become cluttered and unwieldy to manage, particularly if features change which breaks custom solutions. Posting in discussions also invites others to question and iterate, to make improve :)

PR #458, opened 2024-09-17
Fix documentation links

alisonjenkins:
Documenttion links in the README.md were broken by what seems to be a documentation re-organisation. This commit fixes them.


tjex:
So speedy! The plan is to direct people to the new docs site instead of md files in the repo, but I def should've just updated the readme in the mean time as you have. Thank you 🙌

PR #464, opened 2024-10-21
Rename 'ignore' to 'exclude' in default config

smbl64:
New notebooks (`zk init`) use the legacy `ignore` config parameter which does not work anymore.

I believe this was one missed in #322


tjex:
Great catch, thanks :)

PR #469, opened 2024-11-23
build binaries on Actions

jurica:


tjex:
eyyyy! He's back (as promised). That's cool :heart: 

Have you been able to test the build actions somehow? 

Also, there was one entry in the Makefile for Alpine that didn't get deleted. Was it missed, or should it stay in the Makefile for some reason?

jurica:
> ~Have you been able to test the build actions somehow?~ I see this mentioned in the issue thread now 👍
Yeah sorry, I decided to put all relevant information to the issue only to not have to repeat myself all over in the two PRs.

> Also, there was one entry in the Makefile for Alpine that didn't get deleted. Was it missed, or should it stay in the Makefile for some reason?

My idea was to keep the Makefile for local development and I only removed "release related" stuff. So I left the alpine target in there on intention, as it has the additional "-extldflags=-static" flag.

tjex:
Ah yes, of course. One other thing is that the action shouldn't be triggered on push. I'm just shooting this as a quick message, but can make the change later this eve.

And have just seen when updating the branch, that the linux builds failed (as the action ran on push). I haven't looked into why. But will also do so, unless there's an update from your side here. 

Thanks again, this is such a massive workflow optimisation!

jurica:
Can you please post a link to the failed build, I cannot find it…
I wasn‘t exactly sure on the triggers, so I left it as it is and open for discussion. Please go ahead and adjust them as you want. I think a nightly build and build on push of a tag would be sufficient.

tjex:
I'm thinking. Would it be fine to replace the upload-artefact step with [action-gh-release](https://github.com/softprops/action-gh-release?tab=readme-ov-file#%EF%B8%8F-uploading-release-assets) and upload directly to the release?

Or is it better to have a separate workflow for this / add the binaries to the release manually?.. I'm not so versed with this area of github / CI.

jurica:
We could create a release on push of a tag, otherwise upload artifacts. That way a nightly or manual build would be possible and triggering a release would be just a matter of pushing a tag.

tjex:
@jurica How does that look?

I wanted to get the version output from build-binaries to use in release action, so as to reuse variables and not risk mismatches.

So I figured to replace the steps in each build job too. I hope that works. Otherwise should be easily revertable. Tried to make the git commits clean for this.

.github/workflows/build-binaries.yml
```diff
@@ -0,0 +1,181 @@
+name: Build Binaries
+on:
+  workflow_dispatch:
+  workflow_call:
+env:
+  CGO_ENABLED: 1
+jobs:
+  get-version:
```
jurica, line 8: I also tried to get the version just once, but unfortunately it doesn't work. GITHUB_OUTPUT is not shared across jobs, just across steps in a single job...
You would need to write the version into a file in get-version job and store that as an artifact, then pull that artifact in each job again and read the file. Also every job reading the version would need to be dependent on get-version to ensure it is run before. After I figured that out I swallowed the pill of calling git describe in every job.

.github/workflows/release.yml
```diff
@@ -1,19 +1,31 @@
...
-    runs-on: macos-latest
+  build-binaries:
+    uses: ./.github/workflows/build-binaries.yml
+
+  create-release:
+    runs-on: ubuntu-latest
+    needs: [build-binaries]
     steps:
-      - name: Checkout repo
+      - name: Checkout repository
```
jurica, line 14: Why do you need to check out the repo here?

CONTRIBUTING.md
```diff
@@ -73,17 +75,8 @@ and create a new Git version tag (for example `v0.13.0`). Make sure you follow t
...
-chips. The Linux builds are created using Docker and
-[these custom images](https://github.com/zk-org/zk-xcompile), which are hosted via
-[ghcr.io within zk-org](https://github.com/orgs/zk-org/packages/container/package/zk-xcompile).
-
-This process is convoluted because `zk` requires CGO with `mattn/go-sqlite3`, which
-prevents using Go's native cross-compilation. Transitioning to a CGO-free SQLite driver
-such as [cznic/sqlite](https://gitlab.com/cznic/sqlite) could simplify the distribution
-process significantly.
+Binaries can be downloaded from the `build-binaries.yml` Actions run
+that was triggered by pushing the tag and uploaded to the release.
```
jurica, line 79: We need to adjust that as well, I'd keep that open as a last step when we figured out everything else ;).

.github/workflows/release.yml
```diff
@@ -1,19 +1,31 @@
...
         uses: actions/checkout@v4
-      - name: Update Homebrew formula
```
jurica, line 15: This is now missing in the new process, I guess we still need it, right?

.github/workflows/release.yml
tjex: it was in the old release action. Possibly because of the homebrew formula?

CONTRIBUTING.md
tjex: adjusted the whole section a little as it needed more detail anyway I think. But will need to be looked at again later anyway after this pr is sorted

tjex:
@jurica Ok. That should all be implemented. One comment is still unresolved

.github/workflows/release.yml
tjex: true, looks like it doesn't need it: https://github.com/dawidd6/action-homebrew-bump-formula?tab=readme-ov-file#standard-mode

tjex (commented):
So, lets see!

tjex (approved):

tjex:
Working :) https://github.com/zk-org/zk/actions/runs/12073672963/workflow

Two of the linux builds takes ages though! Funny that the alpine build takes the longest :shrug:

.github/workflows/release.yml
jurica: I was about to check it tonight, but you were faster ;).

jurica:
Very nice, I like how that turned out. Thanks for your help and review!

Yeah the arm build for linux run on x86 machines and are emulated with qemu. So they are quite slow.
I was thinking about running them on docker on a mac instance, but then I would have needed to install docker through homebrew first and set everything.
But arm (based runners)[https://github.blog/news-insights/product-news/arm64-on-github-actions-powering-faster-more-efficient-build-systems/] should be available soon, then it's just a matter of adjusting the workflow to execute the arm build on arm runners.

tjex:
ahhh ok. Makes sense re the arm build. 
And no worries, thanks for following through and coming back with the PR. 
Will save us so much effort with releases.

PR #473, opened 2024-11-28
Build action

tjex:
Need to authenticate to ghcr to pull builds in actions


tjex:
Also assigned read rights to the zk repository in zk-xcompile

Issue #482, opened 2024-12-21
--color flag


sivaplaysmC:
### If you have an idea, open a discussion

- [ ] I will [create a new discussion](https://github.com/zk-org/zk/discussions/new?category=ideas) instead of an issue.

sivaplaysmC:
Having a --color flag will be really useful, as it allows piping the output of commands like `zk --list` to other commands like fzf.

sivaplaysmC:
I am willing to implement this feature, and will prepare a pull request asap.

tjex:
closing here as pr in progress

PR #483, opened 2024-12-21
Add `--color` flag

sivaplaysmC:
## TL;DR
Flag `--color` forces colored output, even if stdout is not a terminal.

## Motivation
- `grep --color=always`
- `ls --color=always`

## Application
- Allows colored text when piping to tools like fzf outside of `zk list --interactive`

### DIFF
--- a/main.go
+++ b/main.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 
 	"github.com/alecthomas/kong"
+	"github.com/fatih/color"
 	"github.com/zk-org/zk/internal/cli"
 	"github.com/zk-org/zk/internal/cli/cmd"
 	"github.com/zk-org/zk/internal/core"
@@ -31,9 +32,10 @@ var root struct {
 	Edit  cmd.Edit  `cmd group:"notes" help:"Edit notes matching the given criteria."`
 	Tag   cmd.Tag   `cmd group:"notes" help:"Manage the note tags."`
 
-	NotebookDir string  `type:path placeholder:PATH help:"Turn off notebook auto-discovery and set manually the notebook where commands are run."`
-	WorkingDir  string  `short:W type:path placeholder:PATH help:"Run as if zk was started in <PATH> instead of the current working directory."`
-	NoInput     NoInput `help:"Never prompt or ask for confirmation."`
+	NotebookDir string     `type:path placeholder:PATH help:"Turn off notebook auto-discovery and set manually the notebook where commands are run."`
+	WorkingDir  string     `short:W type:path placeholder:PATH help:"Run as if zk was started in <PATH> instead of the current working directory."`
+	NoInput     NoInput    `help:"Never prompt or ask for confirmation."`
+	Color       ForceColor `help:"Force colored output"`
 	// ForceInput is a debugging flag overriding the default value of interaction prompts.
 	ForceInput string `hidden xor:"input"`
 	Debug      bool   `default:"0" hidden help:"Print a debug stacktrace on SIGINT."`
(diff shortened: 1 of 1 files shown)


sivaplaysmC:
closes #482

tjex:
Nice addition, although using the color flag with `zk list` runs the default list output template and not my own from `config.toml`

`list = 'zk list --format "{{style \"faint\" title}} -> {{style \"blue faint\" path}} ... {{#style \"italic faint\"}}({{format-date created \"elapsed\"}}){{/style}}" $@'
`

This isn't a great experience as each user will be used to having the output formatted in the way they have set it.

Also, once this is sorted, it would be great if you could address the failing test (tesh) cases. Read through CONTRIBUTING.md for clarification on that. I think it's just to do with the standard output from the help menu.

tjex:
So to be clearer, the color force should hook in somewhere deeper, so that it will still be applied to commands run via aliases / modified config settings.

sivaplaysmC:
> the color force should hook in somewhere deeper, so that it will still be applied to commands run via aliases / modified config settings.

ok, but where do you think should the flag be parsed? i think that adding the --color as a part of the alias make more sense. Are we on the same page?

sivaplaysmC:
> Nice addition, although using the color flag with `zk list` runs the default list output template and not my own from `config.toml`
> 
> `list = 'zk list --format "{{style \"faint\" title}} -> {{style \"blue faint\" path}} ... {{#style \"italic faint\"}}({{format-date created \"elapsed\"}}){{/style}}" $@' `
> 
> This isn't a great experience as each user will be used to having the output formatted in the way they have set it.
> 
> Also, once this is sorted, it would be great if you could address the failing test (tesh) cases. Read through CONTRIBUTING.md for clarification on that. I think it's just to do with the standard output from the help menu.

sure, i'll look into it once i complete the implementation 👍🏻

sivaplaysmC:
@tjex , 

> Nice addition, although using the color flag with `zk list` runs the default list output template and not my own from `config.toml`
> 
> `list = 'zk list --format "{{style \"faint\" title}} -> {{style \"blue faint\" path}} ... {{#style \"italic faint\"}}({{format-date created \"elapsed\"}}){{/style}}" $@' `
> 
> This isn't a great experience as each user will be used to having the output formatted in the way they have set it.
> 
> Also, once this is sorted, it would be great if you could address the failing test (tesh) cases. Read through CONTRIBUTING.md for clarification on that. I think it's just to do with the standard output from the help menu.

The purpose of the --color flag is to display colored text, even if the ouput is not a tty (which usually happens when you are piping the output of zk to a pager or fzf).

tl;dr, 
image

image

tjex:
Yep, I'm on the same page. 
So it would seem that this issue should only apply to calling aliases. On line 175 in main.go is the function `runAlias`. Zk executes an alias when it discovers that the first argument of a call to `zk` is in the user's aliases section.

This is why running `zk --color list` for me does not execute my alias.

So the question then is, do we want the `--color` flag to only be available to the `list` command. Based on the current commands we have, it makes sense.

This could be the simplest solution, to just add the `--color` flag as a postional flag for `list`. Then it would simply be appended to an alias.

so `zk <list-as-alias-that-gets-expanded> --color`

Thoughts?

sivaplaysmC:
> do we want the --color flag to only be available to the list command? 
Yes. This actually makes more sense as none of the other commands use colored output (with the exception of `edit -i`, which uses FZF directly).

sivaplaysmC:
> So to be clearer, the color force should hook in somewhere deeper, so that it will still be applied to commands run via aliases / modified config settings.

Ok, can you confirm if you are running the correct zk binary? I just tested it out, and it works perfectly fine.

```bash
export PATH=~/repos/zk:$PATH
cat .zk/config.toml | rg '\[alias' --after-context=2
zk list --color | less
zk --color list | less
```

The last two commands give the same output.

image
image

WhyNotHugo (approved):
~~Not tested, but code changes LGTM.~~

Never mind, comments had not loaded when I commented this.

WhyNotHugo:
Tangential, nonblocking: Do we maybe also want to support NO_COLOR/NO_COLOUR, as in https://no-color.org/?

Issue #253, opened 2022-09-09
Feature Request: Command to list all aliases


gbrls:
I think it'd be a good feature to add a command to list all aliases.

I can work on it some weekend. I'd like to know what you think.

mickael-menu:
Why not, how about `zk config alias` for the command?

tjex:
missclick!

PR #484, opened 2024-12-25
Add `zk config alias`, resolves feature request in issue #253.

andrebauer:


tjex:
Thanks for this :) Having an initial look at it now. May be a bit slow as I'm starting a new job next week.

My initial thoughts is that it would be good to structure this `config.go` file in a way that makes it simple to add new config commands.

In line with that I don't think `zk config` should list out the aliases. It should be `zk config --list aliases` or something to that effect.

There are no doubt future use cases for `zk config ...` commands, so starting out with that in mind will pay dividends down the track.

If you're not up for restructuring yourself, not a worry. I can look at it at some point :+1: 

This is excellent groud work as it is :) Thanks again.

A secondary thought could be to refactor the logic for `cmd.header`, `cmd.delimter`, etc into a function somewhere instead of copying from `cmd.List`. But maybe thats too soon of an optimization.

andrebauer:
If have reworked the `zk config` so that `zk config --list OBJECT` now list the aliases, filters or extras. For further improvements I ask you to give me more specific suggestions. 

Especially for the copied logic from `cmd.List`, the question would be to which file that should be transferred. But this may already be a separate issue?

internal/cli/cmd/config.go
```diff
@@ -0,0 +1,145 @@
...
+	templ, ok := defaultMapFormats[format]
+	if !ok {
+		templ = strings.ExpandWhitespaceLiterals(format)
+	}
+
+	return templ
+}
+
+var defaultMapFormats = map[string]string{
+	"json":  `{"%s":"%s"}`,
```
WhyNotHugo: Use go's native JSON formatting capabilities, this doesn't properly escape items.

andrebauer:
@WhyNotHugo Thanks for your review, I've adjusted my PR.

internal/cli/cmd/config.go
```diff
@@ -0,0 +1,144 @@
...
+	cmd.Header = strings.ExpandWhitespaceLiterals(cmd.Header)
+	cmd.Footer = strings.ExpandWhitespaceLiterals(cmd.Footer)
+	cmd.Delimiter = strings.ExpandWhitespaceLiterals(cmd.Delimiter)
+
+	if cmd.Delimiter0 {
+		if cmd.Delimiter != "\n" {
+			return errors.New("--delimiter and --delimiter0 can't be used together")
+		}
+		if cmd.Header != "" {
+			return errors.New("--footer and --delimiter0 can't be used together")
```
WhyNotHugo:
```suggestion
			return errors.New("--header and --delimiter0 can't be used together")
```

I've seen this block of code repeated elsewhere. We should avoid repeating it, but we can also leave that out-of-scope for this PR.

tjex: Fixed. Yeah, this code block is the same as in `zk list --help`. I had mentioned it [early on](https://github.com/zk-org/zk/pull/484#issuecomment-2569590759), but thought the same -- that it could blow out the complexity of this singular PR.

tjex:
With this approach we can also have an easier way so solve for the config global notebook dir in zk-nvim (https://github.com/zk-org/zk-nvim/discussions/287)

PR #485, opened 2025-01-02
Fix error text when no notes found.

fspeirs:
The error message read "Found 0 note". This patch corrects it to "Found 0 notes.".


tjex (approved):
Fantastic. Thanks for this level of care and attention :100:

PR #487, opened 2025-01-04
Fix typos

kianmeng:
Found via `codespell -S *.svg -L titel,potatoe` and `typos --hidden --format brief`


tjex:
Thanks for the initiative, but these kinds of broad stroke spell corrections have the capacity to cause more problems than they're worth.

PR #491, opened 2025-01-17
use arm based linux runners for arm builds

jurica:
Linux arm runners are [https://github.com/orgs/community/discussions/148648](available), lets use them ;)


tjex (approved):
agreed!

Issue #461, opened 2024-10-05
LSP crashes when adding or removing lines within code fences


tjex:
### Check if applicable

- [x] I have searched the existing issues (**required**)
- [x] I'm willing to help fix the problem and contribute a pull request

### Describe the bug

While editing within code fences, I'm finding that adding lines (i.e, with `<cr>` or `[o|O]`) or removing them (`<BS>` or `dd`) that the LSP crashes with "exit code 2 and signal 0".

Here are the lsp logs:

```text
[START][2024-10-05 18:46:43] LSP logging initiated
[ERROR][2024-10-05 18:46:43] .../vim/lsp/rpc.lua:770	"rpc"	"/Users/tjex/.local/share/nvim/mason/bin/marksman"	"stderr"	"[18:46:43 INF] <LSP Entry> Starting Marksman LSP server: {}\n"
[ERROR][2024-10-05 18:46:52] .../vim/lsp/rpc.lua:770	"rpc"	"zk"	"stderr"	"panic: runtime error: slice bounds out of range [:3] with length 0\n\ngoroutine 25 [running"
[ERROR][2024-10-05 18:46:52] .../vim/lsp/rpc.lua:770	"rpc"	"zk"	"stderr"	"]:\ngithub.com/zk-org/zk/internal/adapter/lsp.isLineWithinCodeBlock({0x1400048c000, 0x17, 0x100cf3000?}, 0x1?, {0x140003ec677, 0x6})\n\t/Users/tjex/.local/src/zk-org/zk/main/internal/adapter/lsp/document.go:222 +0x2c8\ngithub.com/zk-org/zk/internal/adapter/lsp.(*document).DocumentLinks(0x140000423c0)\n\t/Users/tjex/.local/src/zk-org/zk/main/internal/adapter/lsp/document.go:262 +0x10c\ngithub.com/zk-org/zk/internal/adapter/lsp.(*Server).refreshDiagnosticsOfDocument.func1()\n\t/Users/tjex/.local/src/zk-org/zk/main/internal/adapter/lsp/server.go:596 +0x70\ncreated by github.com/zk-org/zk/internal/adapter/lsp.(*Server).refreshDiagnosticsOfDocument in goroutine 23\n\t/Users/tjex/.local/src/zk-org/zk/main/internal/adapter/lsp/server.go:589 +0x10c\n"
```

We worked on code handling links within code fences here, https://github.com/zk-org/zk/pull/399. Perhaps this change introduced a bug.

[...]

PR #493, opened 2025-01-19
Fix LSP Crash in Code Fences

dandeandean:
## Description
This PR fixes the crash described in #461.

The root cause of the problem was that the global state that tracks whether or not a line is in a code fence was not reset after each pass over the file. This behavior can be shown in the `test-wrap.md` file.

This would cause the `currentCodeBlockStart` to not always point to the fence.
Sometimes it would point to a line with len < 3, which resulted in a out of bounds read in `lines[currentCodeBlockStart][:3] == line[:3]`.

Another undefined state caused by this is where `lineIndex` < `currentCodeBlockStart`.

## Testing:

I manually tested by attaching the lsp to `nvim 0v0.10.4` on the following files:
[test-note.md](https://github.com/user-attachments/files/18490023/test-note.md)
[test-fence.md](https://github.com/user-attachments/files/18924318/test-fence.md)
[test-wrap.md](https://github.com/user-attachments/files/18924540/test-wrap.md)

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.yml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yml
@@ -5,7 +5,7 @@ body:
     attributes:
       value: |
         If you have an idea, _please [open as a discussion](https://github.com/zk-org/zk/discussions/new?category=ideas)_ instead of an issue. This helps keep the issue tracker more relavent.
-        Be aware that feature requests [may not get much attention through 2025!](https://github.com/zk-org/zk/discussions/486).
+        Be aware that feature requests [may not get much attention through 2025!](https://github.com/zk-org/zk/discussions/486)
   - type: textarea
     attributes:
       label: Exception
--- a/internal/adapter/lsp/document.go
+++ b/internal/adapter/lsp/document.go
@@ -216,16 +216,23 @@ var currentCodeBlockStart = -1
 // check whether the current line in document is within a fenced or indented
 // code block
 func isLineWithinCodeBlock(lines []string, lineIndex int, line string) bool {
-	if len(line) == 0 { // if 0 either it's in a fence or no
+	if len(line) == 0 {
 		return insideFenced
+	}
+	isEndOfFence := func(line string) bool {
+		if !insideFenced ||
+			currentCodeBlockStart < 0 ||
+			len(line) < 3 ||
+			len(lines) <= currentCodeBlockStart ||
+			len(lines[currentCodeBlockStart]) < 3 ||
+			fencedEndRegex.FindStringIndex(line) == nil {
+			return false
+		}
+		return lines[currentCodeBlockStart][:3] == line[:3]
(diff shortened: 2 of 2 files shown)


dandeandean:
@tjex @Rahlir I am still investigating the root cause of this issue. In the mean time could you take a look at this? I think it may be worth closing this out so there's no crash. Let me know

tjex:
Great, thanks for this!! There's been a lot of activity here all of sudden, and I'm a little down on free time at the moment. So just a heads up that I may take a bit to get around to this.

tjex:
So the test-note.md is causing the crash for me as expected. 

Do you understand why the crash is happening? 

I feel like it would be great to just hit the nail on the head with this from the start, instead of a hot fix.
Because this will likely become an "out of sight out of mind" situation. And then some time down the track we'll forget the context of the code you've provided, and that it should in fact be removed in place of a more robust solution.

dandeandean:
@tjex That's a fair point. I haven't been working on this, but I can pick up on this again & try to find a more robust solution. I was being lazy & trying to get away with just fixing the crashing conditions. 

It's strange that it's still crashing for you... how did you test it?

tjex:
Ah sorry. I meant that I tested the functionality of the test-note to recreate a crash. And that worked. 

If you wanted to do so that would be awesome. I can help along the way. But sporadically... Still getting to grips with the new job and balancing a masters thesis on the side.

But this bug is a high priority for me.

dandeandean:
@tjex I've been looking this week trying to find out what causes these crashing conditions, but no dice. The problematic function is `isLineWithinCodeBlock`, which is called by `DocumentLinks`. `DocumentLinks` is called in line 596 in the server (which is a goroutine). We can see all of this in the stack trace. 

We also call this `DocumentLinks` in `handler.TextDocumentDocumentLink`. I suspect our issue comes from that function calling `DocumentLinks` unexpectedly & messing up the expected state of the global variables that we keep track of fence information. It may be worth redesigning how this `isLineWithinCodeBlocks` works so it doesn't have side-effects. 

Do you have any insight into when this `handler.TextDocumentDocumentLink` is called? I can't seem to find any documentation on `glsp16`.

dandeandean:
Hey @tjex, after some more investigation (and a whole bunch of print debugging), I am pretty sure the fix was way more simple than I initially though. See the updated the description & the test files. 

After compiling & attaching this version of the LSP, I haven't had any crashes.
If you could independently verify, that'd be great, but I think this is ready to merge.

tjex:
Ooooooh dayum. That's a nice turn of events! Will check it out after dinner / tomorrow. 
Thank you!!! This is an awesome result (if it does indeed do the trick haha). Thank you for your persistence!

tjex:
All working on my end. Well done and thanks again! 
Really happy about this 🥳

tjex (approved):

Issue #490, opened 2025-01-13
Overlapping group paths confusion


dodgog:
### Check if applicable

- [X] I have searched the existing issues (**required**)
- [X] I'm willing to help fix the problem and contribute a pull request

### Describe the bug

When one group defines a folder for its path, and another group defines as its path the subfolder of the folder, conflicts arise on creation of the note in the subfolder. It seems to randomly decide which group to assign the note to. 50% into folder's group, 50% into subfolders group.

### How to reproduce?

zk new --no-input "$ZK_NOTEBOOK_DIR/area/subfolder/"
zk new --no-input "$ZK_NOTEBOOK_DIR/area/subfolder/"

50% of the time this will create a notebook with the template 1 and 50% with the template 2.

### zk configuration

```toml
[group.area]
paths = ["area"]
[group.area.note]
extension = "md"
template = "template1.md"

[group.subarea]
paths = ["area/subfolder"]
[group.subarea.note]
extension = "md"
template = "template2.md"
```

### Environment

```bash
zk 0.14.1
system: Darwin 23.1.0 arm64
```

tjex:
Yikes... Confirmed.

tjex:
closed with #495

PR #495, opened 2025-01-20
fix: nested group rules

dodgog:
An attempt to fix the bug described in https://github.com/zk-org/zk/issues/490

The bug was due to early returning when a partial prefix match happened.
We now find the longest partial prefix match and return that instead.

### DIFF
--- a/internal/core/config.go
+++ b/internal/core/config.go
@@ -113,8 +113,8 @@ func (c Config) GroupConfigNamed(name string) (GroupConfig, error) {
 // GroupNameForPath returns the name of the GroupConfig matching the given
 // path, relative to the notebook.
 func (c Config) GroupNameForPath(path string) (string, error) {
-    var longestMatch string
-    var matchedName string
+	var longestMatch string
+	var matchedName string
 
 	for name, config := range c.Groups {
 		for _, groupPath := range config.Paths {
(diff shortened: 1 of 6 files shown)


dodgog:
I apologize, I'm still trying to get tesh to work. Let me know if it's necessary please, or if the current tests suffice.

dodgog:
I apologize in advance if I messed something up in the pr, I would be happy to correct or redo – it's my first time contributing here.

tjex:
Thanks for this.

Having a quick first look now.

tjex:
So I found an edge case that breaks this solution.

```toml
[group.t1]
paths = ["level1", "level1/level2/level3"]
[group.t1.note]
extension = "md"
template = "template1.md"
filename = "template1-{{id}}"

[group.t2]
paths = ["level1/level2"]
[group.t2.note]
extension = "md"
template = "template2.md"
filename = "template2-{{id}}"
```

So if a different group has a path between the paths of another group, it fails.

In the case above, `zk new level1/level2` will still create a note with template 1, instead of template 2.

I think the way to go will be to apply group rules explicitly to each defined path. Any child directories that are not explicitly defined, will then adopt the rules of the parent directory.

This way, `level1/level2` would trigger template2, as it has a group rule defined. If it _didn't_ then it would adopt the rule of `level1`.

While this is a bit of an edge case, and your solution does fix the current issue, I think we're very close to making a much more robust solution. 

I'll also have a think of how to solve this. 
Feel free to write me here if you want to go over anything together 👍

tjex:
One approach could be to create a map from all group paths. 
Then on the creation of a new note, check if the path exists in the map, if so return the group name (the value of the key. Where the key is the group path).

If none is found, recurse upwards until notebook root.

This would also give us the ability to return an error if there are two group rules for a singular path (i.e, a conflict).

tjex (approved):

PR #506, opened 2025-02-04
readme(nix): Don't recommend `nix-env -iA`

srid:
https://stop-using-nix-env.privatevoid.net/


tjex:
But this is in the case that installing a package installs other dependent packages, which may overwrite (and change versions of) those packages right?

In our case, we're just installing one binary.

worldmind:
Trying to install it:
```
$ nix-env -iA zk
error: attribute 'zk' in selection path 'zk' not found
```
although
```
nix run nixpkgs#zk
```
works, perhaps something changed
```
$ nix --version
nix (Nix) 2.26.3
```

worldmind:
```
nix profile install nixpkgs#zk
```
works

worldmind:
also `nix profile install nixpkgs#fzf` is needed for use interactive mode

tjex:
@worldmind would you be able to consult on what the best option is for the readme? I don't have the time to understand the quirks of a new distro right now. 

A PR would of course be best, or you can format it as you would here in thread in markdown and i can apply it.

worldmind:
Sorry, but I am afraid I know nothing, I installed nix on Ubuntu just for install zk, and can't say I got what is the right way, somehow I found working way and decided to not go deeper as for now I am not sure I will use `nix` actively (perhaps next time I will use it via docker).
I disabled warnings:
```
$ cat .config/nix/nix.conf 
experimental-features = nix-command flakes
```
and after executing:
```
nix profile install nixpkgs#zk
nix profile install nixpkgs#fzf
```
it works.

testplayername:
> But this is in the case that installing a package installs other dependent packages, which may overwrite (and change versions of) those packages right?
> 
> In our case, we're just installing one binary.

Nix is specifically designed to avoid the problem of [dependency hell](https://en.wikipedia.org/wiki/Dependency_hell). This is never an issue when using Nix.

---

Nix is a complex beast, with a package management system that looks alien compared to other package managers. A comment on a Github issue is a poor place to explain all that is Nix, so I'll try to provide a short summary of Nix here, and just the parts that are immediately relevant to this issue.

One of the primary points of Nix is _declarative_ package management and configuration. Whereas with other package managers the way to install something would be a shell command, e.g. `sudo apt install firefox` and `sudo pacman -S firefox`, the best way to install a package in Nix is _editing configuration files_. This is conceptually similar to the `go.mod` for Go programs, but in the case of Nix, these configuration files defines a configuration of a system, including what programs are installed.

However, Nix also provides several other ways to "install" a package. This is where `nix-env -iA zk` and `nix profile install nixpkgs#zk` comes from. These two commands install the package _imperatively_, and I would argue both of these commands are anti-patterns, because Nix is designed for declarative package management.

Nix (and NixOS) configurations can get highly customized. Installation instructions necessarily have to assume the user has some basic Nix knowledge. I recommend these lines in the section for Nix in the readme:

---
### Nix
`zk` is available in nixpkgs and has a [Home Manager](https://github.com/nix-community/home-manager) module.

If you want to run `zk` without permanently installing it:
```
nix run nixpkgs#zk
```
Or, if you want to create an ephemeral shell with `zk` available:
```
nix shell nixpkgs#zk
```
To permanently install `zk` on NixOS, include `nixpkgs.zk` in `environment.systemPackages` in your system configuration (`/etc/nixos/configuration.nix` by default).

```
environment.systemPackages = [
  # Your other packages here
  nixpkgs.zk
];
```
If you are using [Home Manager](https://github.com/nix-community/home-manager), you can permanently install and configure `zk` just for your user via the Home Manager module. Add this to your Home Manager configuration:
```
programs.zk.enable = true;
```

tjex:
Perfect. Thank you for the concise explanation and the text, which I think is perfect 👌

@srid , can you update the pr with the above?

testplayername:
Made an alternative PR with my changes in #538.

Issue #518, opened 2025-02-13
Ctrl+E does no longer create new notes in chooser


pinpox:
### Check if applicable

- [x] I have searched the existing issues (**required**)
- [x] I'm willing to help fix the problem and contribute a pull request

### Describe the bug

When in the note chooser (`zk edit --interactive`) pressing <kbd>Ctrl</kbd>+<kbd>e</kbd> used to create a new note with the entered query as title. Now it only exits zk.

### How to reproduce?

1. Run `zk edit --interactive`
2. Type something
3. Press Ctrl+e
4. zk exits instead of creating a new note

### zk configuration

```toml
# vim: filetype=dosini
# zk configuration file
#
# Uncomment the properties you want to customize.

# NOTE SETTINGS
#
# Defines the default options used when generating new notes.
[note]

# Language used when writing notes.
# This is used to generate slugs or with date formats.
language = "en"

# The default title used for new note, if no `--title` flag is provided.
default-title = "Untitled"

# Template used to generate a note's filename, without extension.
filename = "{{format-date now '%Y-%m-%d-%H%M%S' }}"

# The file extension used for the notes.
extension = "md"

# Template used to generate a note's content.
# If not an absolute path, it is relative to .zk/templates/
template = "default.md"

# Configure random ID generation.

# The charset used for random IDs. You can use:
#   * letters: only letters from a to z.
#   * numbers: 0 to 9
#   * alphanum: letters + numbers
#   * hex: hexadecimal, from a to f and 0 to 9
#   * custom string: will use any character from the provided value
#id-charset = "alphanum"

# Length of the generated IDs.
#id-length = 4

# Letter case for the random IDs, among lower, upper or mixed.
#id-case = "lower"


# EXTRA VARIABLES
#
# A dictionary of variables you can use for any custom values when generating
# new notes. They are accessible in templates with {{extra.<key>}}
[extra]

#key = "value"


# GROUP OVERRIDES
#
# You can override global settings from [note] and [extra] for a particular
# group of notes by declaring a [group."<name>"] section.
#
# Specify the list of directories which will automatically belong to the group
# with the optional `paths` property.
#
# Omitting `paths` is equivalent to providing a single path equal to the name of
# the group. This can be useful to quickly declare a group by the name of the
# directory it applies to.

#[group."<NAME>"]
#paths = ["<DIR1>", "<DIR2>"]
#[group."<NAME>".note]
#filename = "{{date now}}"
#[group."<NAME>".extra]
#key = "value"

[...]

PR #526, opened 2025-02-25
fix: Ctrl+E does no longer create new notes in chooser

gyorb:
There was a --bind behavior change in fzf 0.53.0 which could be causing the error.

execute action is for executing an external program in the alternate screen without leaving fzf. In zk abort+execute was used to close fzf and open a zk instance.

become action is similar to execute, but it replaces the current fzf process with the specified command using execve(2) system call.

This should have similar behavior to the previous abort+execute action.

https://github.com/junegunn/fzf/issues/3865  
https://github.com/junegunn/fzf/blob/710ebdf9c14e0c88b0cfc843ce08edcdd4554571/man/man1/fzf.1#L1785-L1789  
https://github.com/junegunn/fzf/commit/6ea38b44384e7a09a3863465dc3cc7b93cd7e781

fixes: #518


mcDevnagh (approved):

mcDevnagh:
tested and looks good. Thank you!

PR #527, opened 2025-02-28
Use doublestar for group paths

mcDevnagh:
As of right now, there's no way to match _across file separators_. For example, if I would like to match both `apple/ambrosia` and `melon/cantaloupe/ambrosia`, I would need to specify both paths:
```toml
[group."Ambrosia Fruit Varieties"]
paths = ["apple/ambrosia", "melon/cantaloupe/ambrosia"]
```
Note that `*/ambrosia` does __not__ work, since it doesn't match across `/`; it only matches `apple/ambrosia`
I would prefer to be able to specify just the `**/ambrosia` once, especially if I were a botanist trying to develop a highbush ambrosia blueberry that I would file it under `berry/blueberry/highbush/ambrosia`

We already use [doublestar](https://github.com/bmatcuk/doublestar) when looking for [excluded paths](https://github.com/zk-org/zk/blob/f4d3dc7113f5639fb262cb2d342419a8a1249814/internal/core/note_index.go#L140), so this doesn't even add a new dependency. I also added a couple tests, including over regular `*` globs.

### DIFF
--- a/docs/config/config-group.md
+++ b/docs/config/config-group.md
@@ -21,19 +21,18 @@ paths = [
 ```
 
 You can also use
-[glob patterns](https://en.wikipedia.org/wiki/Glob_(programming)) in `paths`.
+[glob patterns](https://en.wikipedia.org/wiki/Glob_(programming)) in `paths`:
 
 ```toml
 [group.journal]
+# Apply to child directories of "journal".
 paths = ["journal/*"]
-```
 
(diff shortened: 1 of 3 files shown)


tjex (approved):
Fantastic idea. Tested and working. 
Added it to the documentation too. 

Thank you and welcome back :)

PR #531, opened 2025-03-27
add `creation-date-key` `modification-date-key` configuration options

BennoLossin:
These two configuration options are used to change the way `zk` obtains the creation date and modification date from a file.

I have a lot of notes that use `created` and `changed` as the keys for the creation and modification date (I store them in git and thus the true creation/modification date are useless).

This also is useful when one wants to use a different language for the keys in the frontmatter.

I am pretty new to go and this is my first contribution to `zk`, so I very likely have missed something.


y86-dev:
Ah I just noticed that I didn't finish the `tesh` files correctly... I'll get to that tomorrow, never used `tesh` before and the `make tesh` command didn't result in any errors, so I didn't know what the correct output should be.

BennoLossin:
The motivation is the following: I'm using a git repository to store my notes & sync them between multiple devices. This means that things like `zk edit -i --sort=modified` don't reflect the actual modification date when I last touched the file, but when `git` last touched it. That's why the new `changed` yaml key is needed. Since we now have two dates & times in the yaml frontmatter, it didn't really sound like a good idea to have "`date`" as a key, since it could lead to confusion. Additionally, I use the `created` key in my notes before I switched to zk. Thus allow users to specify a custom name for the creation date key in the yaml frontmatter.

The user can specify `format.markdown.frontmatter.creation-date-key` (default value is `"date"`), this will replace the existing `date` key and all logic that uses it. Similarly, the `format.markdown.frontmatter.modification-date-key` option will allow users to use this key in frontmatter to specify the actual modification date that will take priority over the files `mtime`. It will be used for things like `zk list --sort=modified`.

BennoLossin:
Indeed, I have set up my editor to do that for me. Other ways to do it are for example git hooks.

tjex:
There is now also an effort to auto update the modification date at zk-org/zk-nvim#272

But I'd prefer to support this on the side of the `zk` binary, and not zk-nvim

internal/core/config.go
```diff
@@ -161,6 +165,18 @@ type MarkdownConfig struct {
...
 	// Indicates whether a link's path file extension will be removed.
 	LinkDropExtension bool
+
+	// Frontmatter determines the keys used in the frontmatter
+	Frontmatter YamlFrontmatterConfig
+}
+
+// YamlFrontmatterConfig holds the configuration for Yaml frontmatter.
+type YamlFrontmatterConfig struct {
+	// CreationDate indicates what name the key for the creation date has. Default is "date"
```
WhyNotHugo:
```suggestion
	// CreationDate is the key for the creation date has. Default is "date"
```

internal/core/config.go
```diff
@@ -161,6 +165,18 @@ type MarkdownConfig struct {
...
+	CreationDate     string
+	// ModificationDate indicates what name the key for the modification date has. If not present,
```
WhyNotHugo:
```suggestion
	// ModificationDate is the key for the modification date has. If not present,
```

WhyNotHugo (approved):
LGTM, I only have very minor suggestions wrt reducing verbosity.

BennoLossin:
Sorry for the long wait, I was very busy in the last two months.

I added a lot more docs now and also included automation for neovim and git. Would probably be a good idea to also include other common editors like vscode, emacs etc. But I'm not familiar with those, so I'd leave that to others. (I also rebased onto the lastest changes from `main`)

One thing that also could be done would be to add the `changed` key to the default configuration i.e. no config changes are needed to make `changed: <date>` work.

tjex:
Fantastic. Thanks for getting back on this <3

docs/tips/modification-date.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Configuration
+
+First, configure `zk` to use the `changed` frontmatter key to obtain the
+modification date in `.zk/config.toml`:
+
+```toml
+[format.markdown.frontmatter]
+modification-date-key = "changed"
+```
```
WhyNotHugo, line 17: Should link to note-frontmatter.md / Date Keys.

docs/tips/modification-date.md
```diff
@@ -0,0 +1,85 @@
...
+
+for file in $(git diff --cached --name-only -- '*.md'); do
+  if [ -f "$file" ] && grep -q '^changed:' "$file"; then
+    sed -i. bak "s/^changed:.*/changed: $(date +'%Y-%m-%d %H:%M:%S')/" "$file"
+    rm -f "${file}.bak"
+    git add "$file"
+  fi
+done
+
+#!/usr/bin/env bash
```
WhyNotHugo: Hashbang in middle of script is erroneous:
```suggestion
```

docs/tips/modification-date.md
```diff
@@ -0,0 +1,85 @@
...
+})
+```
+
+### Git Pre-commit Hook
+
+Create a Git pre-commit hook to automatically update modification dates before committing.
+Save this as `.git/hooks/pre-commit` in your notebook repository:
+
+```bash
+#!/bin/bash
```
WhyNotHugo: This is a plain shell script with nothing bash-specific:

```suggestion
```sh
#!/bin/sh
```

tests/config-format-markdown-frontmatter-creation-date.tesh
```diff
@@ -0,0 +1,7 @@
+$ cd blank
+
+$ echo "[format.markdown.frontmatter] creation-date-key = \"created\"" > .zk/config.toml
+
+$ echo "---\ncreated: 2025-03-27 01:18:43\n---\n\n# Note" > note.md
+$ zk list -qflink
+>{"filename":"note"}
```
WhyNotHugo: How is this testing anything related to creation date?

tests/config-format-markdown-frontmatter-modification-date.tesh
```diff
@@ -0,0 +1,7 @@
+$ cd blank
+
+$ echo "[format.markdown.frontmatter] modification-date-key = \"changed\"" > .zk/config.toml
+
+$ echo "---\nchanged: 2025-03-27 01:18:43\n---\n\n# Note" > note.md
+$ zk list -qflink
+>{"filename":"note"}
```
WhyNotHugo: How is this testing anything related to modification date?

WhyNotHugo (changes requested):

docs/tips/modification-date.md
BennoLossin: seems like I left in a previous script, fixed now

tests/config-format-markdown-frontmatter-creation-date.tesh
BennoLossin: I have forgotten why I wrote these tests, I removed them, feel free to suggest correct ones :)

God-damnit-all:
I'm really looking forward to this being implemented, the lack of integrated date-tracking is the main reason why I haven't made zk my primary note-taking app.

tests/config-format-markdown-frontmatter-creation-date.tesh
tjex: A tesh test would need to execute with filechanges in order to test this functionlity. 

So `zk new` and check the creation date.

Not sure how the modified-date would be tested though, as a note would have to be opened by zk, edited and then saved..

WhyNotHugo:
On Mon, 26 Jan 2026, at 16:07, Tillman Jex wrote:
> 
> 
> ***@***.**** commented on this pull request.
> 
> 
> In tests/config-format-markdown-frontmatter-creation-date.tesh <https://github.com/zk-org/zk/pull/531#discussion_r2727978959>:
> 
> > +$ echo "---\ncreated: 2025-03-27 01:18:43\n---\n\n# Note" > note.md
> +$ zk list -qflink
> +>{"filename":"note"}
> A tesh test would need to execute with filechanges in order to test this functionlity.
> 
> So `zk new` and check the creation date.
> 
> Not sure how the modified-date would be tested though, as a note would have to be opened by zk, edited and then saved..
> 

I think you can use touch to update the notes mtime and then `zk index`.

You can also set EDITOR=touch I guess?

BennoLossin:
Sorry, I won't have the time to add tests myself, if any of you want to add them, feel free to send them to me or just directly push them if that's possible.

tjex:
@BennoLossin , just to confirm.
I've tried to add what should be a [functioning test fixture](https://github.com/zk-org/zk/pull/531/commits/667aebe78135b1b2f16baee6c3d6b4e98751d761). 
By my understanding only the note without a custom changed value should be returned?
In any case, the note that has a modification date _prior_ to the query, should not be returned. 

Run in `test/fixtures/creation-date-key`

```
$ -> zk list --modified-after=2026-01-03
Custom Changed Date ... 2 minutes ago
No Changed Date ... 2 minutes ago
```

BennoLossin:
I'm not sure how the parser handles an empty value. What happens when you remove the key `changed: ` in the "No Changed Date" note entirely?

Edit: if it's empty, I guess it returns `times.ModTime().UTC()`, so the mtime of the file

tjex:
That's a potential secondary issue regarding an empty `changed:` key. The main issue is that the note with a changed date was set to `2026-01-01 19:00` and should not be returned by `zk list --modified-after=2026-01-03`

tjex:
Looking at this again from my end.

tjex:
So I think this is all good now. Regarding testing the modification date. I'm not sure we were thinking about it in the right way. 

Creating a new note with `zk new` in the tesh run, will have the mtime of the note set to the time of creation by default. So we can test this simply by creating a note.

To test the actual functionality of this PR (a custom mod time), we just have a saved note with a custom modified time in the past.

Same logic for testing the creation key.

PR #533, opened 2025-03-30
Fix completion of markdown links with LSP titles

### LINKED ISSUE #435: Titles inside markdown links breaks LSP link completion
### Check if applicable

- [X] I have searched the existing issues (**required**)
- [X] I'm willing to help fix the problem and contribute a pull request

### Describe the bug

Related to #436 

As was uncovered during discussion here: https://github.com/zk-org/zk-nvim/discussions/176#discussioncomment-10054556

The trigger pattern outlined [here](https://github.com/zk-org/zk/blob/725126575d00130c2fa423d36c89f0fff328a318/internal/adapter/lsp/server.go#L664-L665) for regular markdown links is not triggering lsp completion properly.

We have previously looked at a similar issue with wikilink completion: [zk-org/zk#397](https://github.com/zk-org/zk/pull/397)

### How to reproduce?

1. open a note
2. type `[link]((` 

LSP completion should trigger.

### zk configuration

```toml
n/a. (although `link-format` in config will define how the link is subsequently inserted).
```

### Environment

```bash
zk 0.14.1
system: Darwin 23.4.0 arm64
```

vandalt:
Fixes #435 and zk-org/zk-nvim#176.

After looking at the code and testing earlier versions, it seems that the issue was introduced in #397, which fixed the issue of triggering completion manually, but also introduced offsets that account for words instead of just using a `-2` offset. The `-2` offset worked for both `[[` and `((`, but the checks in #397 only accounted for `[[` which broke completion for `((`.

This PR just adds an extra condition to also catch `((`. Some notes:

- I don't really know go, so comments welcome
- It seems that trigger (non-auto) completion with  `[[|` was fixed by #397, but that `[[somewords|` still does not work. And it still does not work with these small updates (but `((|` works just like `[[`)
- I'm marking this as a draft for now because I want to double check the TODO comments I added (basically `((` is included in the word regex, but `[[` is not, so I need to account for that

### DIFF
--- a/internal/adapter/lsp/server.go
+++ b/internal/adapter/lsp/server.go
@@ -856,7 +856,7 @@ func (s *Server) newTextEditForLink(notebook *core.Notebook, note core.MinimalNo
 			startOffset = -2
 		} else {
 			currentWord := doc.WordAt(pos)
-			startOffset = -2 - len(currentWord) 
+			startOffset = -2 - len(currentWord)
 		}
 	}
 
--- a/internal/adapter/lsp/server.go
+++ b/internal/adapter/lsp/server.go
@@ -644,7 +644,8 @@ func (s *Server) refreshDiagnosticsOfDocument(doc *document, notify glsp.NotifyF
 // completion started automatically when typing an identifier, or manually.
 func (s *Server) buildInvokedCompletionList(notebook *core.Notebook, doc *document, position protocol.Position) ([]protocol.CompletionItem, error) {
 	currentWord := doc.WordAt(position)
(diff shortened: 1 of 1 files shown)


vandalt:
Ok I think this is ready for review. The second bullet point above is also fixed: invoked (non-auto) completion now also works with parentheses.

Thanks!

tjex:
Thanks for all of this, and sorry for being slow. I'm juggling a lot atm, but will cool off round July / August

tjex (approved):
Tested with and without autopairs enabled. All working and code changes seem straight forward enough, thanks!

PR #540, opened 2025-04-22
ci: change tar command on macOS runners

williamboman:
There seems to be some weird behaviour going on when creating compressed
tarballs with `bsdtar` this way (redirecting stdout). The produced files
seems OK, but `gzip` itself struggles to decompress it without errors,
resulting in a "trailing garbage ignored" error message and exit code 2
(it still decompresses the file just fine though). This error doesn't
surface when decompressing using `bsdtar` directly, but it does when
using `gnutar`.

See https://github.com/mason-org/mason-registry/pull/9775.

If you have a Mac you should be able to reproduce this assuming `tar ->
bsdtar` and that you have `gnutar` installed as `gtar`.

```
$ echo Hello > file
$ tar -cvz file > archive.tar.gz
$ gzip -d -c archive.tar.gz > /dev/null

$ tar -cv file > archive.tar
$ gzip archive.tar
$ gzip -d -c archive.tar.gz > /dev/null

$ tar -czf archive.tar.gz file
$ gzip -d -c archive.tar.gz > /dev/null
```

I figured I'd open a PR since I found this somewhat interesting and
spent some time debugging it. Not entirely sure I've even found the root
cause but I believe this should help.


tjex:
Really appreciate this! Very considerate and out of the blue :) 

Will also take the opportunity to extend a big thank you for Mason 🙏

tjex (approved):

williamboman:
Thanks! Would it be possible to get a new patch release with new assets? There's some problems installing `zk` on Mac at the moment via Mason

tjex:
Done: https://github.com/zk-org/zk/releases/tag/v0.15.1

PR #542, opened 2025-04-30
Add crossplatform home directory detection

felixhudson:
I have noticed that the global config is not being used on windows. Here is my attempt to fix it.

### DIFF
--- a/internal/cli/container.go
+++ b/internal/cli/container.go
@@ -176,16 +176,15 @@ func globalConfigDir() string {
 	if !ok {
 		home, ok := os.LookupEnv("HOME")
 		if !ok {
-      var err error
-      home, err = os.UserHomeDir()
-      // and fall back to unix default
-      if err != nil {
-        home = "~/"
-      }
+		      var err error
+		      home, err = os.UserHomeDir()
+		      // and fall back to unix default
+		      if err != nil {
+			      home = "~/"
+		      }
 		}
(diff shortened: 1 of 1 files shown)


tjex:
Makes sense, but you shouldn't need to create an err variable?
The method returns an error.  https://pkg.go.dev/os#UserHomeDir

felixhudson:
Yeah its odd having creating the error variable, but without it the compiler complains that err doesnt exist. Or with the := method it claims I can't override the home variable from the parent scope. 

An alternative is to have  home2, err := os.UserHomeDir() , then have an extra if statement to compare home with home2 and create a path with the correct one. But I thought that is just introducing an extra variable for no reason, and adds extra complexity.

I found this solution to be the most concise.

tjex:
It occured to me that we have a path resolver util function. Could you try `ExpandPath(path)` from `internal/util/paths/paths.go`?

felixhudson:
I have reworked all of the home detection logic. I have tested on windows and linux.

tjex:
Haven't you just rewritten the same logic as before? Or am I mistaken? Bascially, you're wanting to get to `os.UserHomeDir()` to support Windows. But that is already theoretically happening in the original code.

`$XDG_CONFIG_HOME` not `$HOME` will exist in windows, so the code will reach the line in the if statement to set home to `os.UserHomeDir()`.

PR #556, opened 2025-07-04
Include version in release tarballs

WhyNotHugo:
Configure git to include the current version in tarballs generated via git-archive.

Fixes: https://github.com/zk-org/zk/issues/555
See: https://whynothugo.nl/journal/2024/12/16/injecting-project-versions-into-builds/

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+VERSION.txt export-subst
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,7 +22,7 @@ CGO_ENABLED=1 GOARCH=arm64 go build -tags "fts5" -ldflags "-X=main.Version=`git 
 - `GOARCH=arm64` is only required for Apple Silicon chips.
 - `-tags "fts5"` enables the FTS option with `mattn/go-sqlite3`, which handles
   much of the magic behind `zk`'s `--match` filtering option.
-- ``-ldflags "-X=main.Version=`git describe --tags --match v[0-9]* 2> /dev/null` -X=main.Build=`git rev-parse --short HEAD`"``
+- ``-ldflags "-X=main.Version=`git describe --tags --match v[0-9]* 2> /dev/null`"``
   will automatically set `zk`'s build and version numbers using the latest Git
   tag and commit SHA.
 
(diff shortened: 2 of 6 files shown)


tjex:
I'm not getting the semver portion of the version number. And the archived src is also complaining.

```
~/.local/src/zk-hugo version
0 > make
CGO_ENABLED=1 go build -tags "fts5" -ldflags "-X=main.Version=79bdb53 -X=main.Build=`git rev-parse --short HEAD`"

~/.local/src/zk-hugo version
0 > ./zk --version
zk 79bdb53

~/.local/src/zk-hugo version
0 > git archive HEAD > ~/tmp/zk.tar
cd ~/tmp/
tar xf zk.tar
make
./zk --version
CGO_ENABLED=1 go build -tags "fts5" -ldflags "-X=main.Version= -X=main.Build=`git rev-parse --short HEAD`"
fatal: not a git repository (or any of the parent directories): .git
zk
```

I'm wondering if it's simpler to just write out the version to VERSION.txt either automatically on tag creation or otherwise. And then just have the makefile build/install inject it via ldflags?

tjex:
This works, where VERSION is a file at root containing `0.15.1` / `<actual version>`.
```
ifneq ("$(wildcard .git)","")
    VERSION := $(shell git describe --tags --match v[0-9]* 2> /dev/null)
    BUILD := $(shell git rev-parse --short HEAD)
else
    VERSION := $(shell cat VERSION)
    BUILD := release
endif
```
We could write that file in the release workflow, so there need not be any extra manual steps on the developer's end for releaseing.

Regarding the tesh test... I'd either write some straight text in the output, that if you're running tesh from a tar, then this will fail. Or just leave it in. Therer shouldn't be so many people developing from a tar ball? And if they are, they're probably not going to contribute back up stream, because they need to then push anyway -- whereby they'll need the .git directory and its contents/history.

WhyNotHugo:
I removed `BUILD := git rev-parse --short HEAD` since the commit is always included when **not** building from a tag. E.g.:

```
> ./zk --version
zk 0.15.1-9-g953893f
```

When building from a tag, only the tag is included:

```console
> git tag v99
> make
> ./zk --version
zk 99
```

tjex:
Yup, hadn't pulled the tags from upstream. That's all tested and working on my end. 

I got the tesh case to work in the tar and in the repo. Was a small change, so just pushed straight here. Let me know what you think, and if you can ideally test on your end, would be ace.

WhyNotHugo:
Changes to tests look fine to me.

CI failures are unrelated?

PR #570, opened 2025-09-07
fetch flags to support tesh cases

tjex:
flag-version.tesh is failing in ci. The output shows that the version is not being read, only the commit ref. As was pointed out to me before in #556 , git tags need to be present in order to get that version info. So hopefully that fixes the failing tesh case here too.


tjex:
fixed in #592

Issue #569, opened 2025-09-07
Tesh cases began failing with bash


tjex:
### Check if applicable

- [x] I have searched the existing issues (**required**)
- [ ] I'm willing to help fix the problem and contribute a pull request

### Describe the bug

Since dependabot (and I) bumped workflow versions, CI started failing on the tesh cases. E.g, 
https://github.com/zk-org/zk/actions/runs/17526821338/job/49778383865

But even after reverting these changes in #568 , CI is still failing. 

### How to reproduce?

Run CI workflow `build`.

tjex:
It would seem that at some point the github bash runner started acting as it should have  been the whole time?.. Many of the tesh cases have lines such as `echo "[note]\n exclude = ['drafts/**']"`.

In bash, this results in a literal string: `[note]\n exclude = ['drafts/**']"`

`printf` needs to be used to print the control characters (new lines).

Or we change the shell runner to zsh. Does anyone even use bash as their main shell anymore?...

WhyNotHugo:
I tried looking into this locally but had a really hard time.

When I run tests with 24569f9e110f889cbdd7f9b5ef7cd33709bcf6a9 (latest master) a few tests fail.

When I run test with f477aacf0dd2fb1812529fc1e54456d36ce5d15e (the last GREEN commit in master) I still get some test failures.

This ruled out bisection for me, but I also can't understand why commits that pass in CI don't pass for me.

My default interactive shell in `zsh`, but `/bin/sh` points to busybox's ash.

WhyNotHugo:
My `$SHELL` is `/bin/zsh`, and the following tests fail for me (including on the version that was green on CI):

```
FAIL cmd-new.tesh: $ cat dry-run.md
expected on stderr:
---
cat: dry-run.md: No such file or directory

---
got:
---
cat: can't open 'dry-run.md': No such file or directory

---
```

And:

```
FAIL cmd-list-filter-date.tesh: $ zk list -qf\{{title}} --created today
expected on stdout:
---
Buy low, sell high
Channel
Compound interests make you rich
Concurrency in Rust
Dangling pointers
Data race error
Diversify your portfolio
Do not communicate by sharing memory; instead, share memory by communicating
Don't speculate
Errors should be handled differently in an application versus a library
Fearless concurrency
Financial markets are random
Green threads
How to choose a broker?
Investment business is a scam
Message passing
Mutex
Null references: the billion dollar mistake
Ownership in Rust
Stick to your portfolio strategy
[...]

PR #575, opened 2025-09-16
feat: respect ZK_CONFIG_DIR as an override for global config location and update docs

Alb-O:
Always a nicety when programs provide an option like this. For my own usecase, I prefer to wrap config files into the nix store and point the program at it.  This variable adds this flexibility for zk.


tjex:
I love me a custom ENV variables :) Tested `tesh` cases locally. They're broken in CI, and that's known: #569

PR #577, opened 2025-09-17
Add LSP diagnostic for missing backlinks

WhyNotHugo:
Implements a new diagnostic that warns when other notes link to the current note but the current note doesn't link back to them.

Use indexed data to find incoming links. Check current document content (instead of the index) for outgoing links to ensure the diagnostics disappears/reappears while editing as appropriate).

By default, this diagnostic is disabled.


WhyNotHugo:
I've also implemented a code action to insert missing back-links, although that likely requires more work and is best reviewed after this is merged: https://github.com/WhyNotHugo/zk/commit/f7d9cb641fb27330f6b5746176ae646954c5f93b

WhyNotHugo:
On Wed, 1 Oct 2025, at 15:21, Tillman Jex wrote:
> *tjex* left a comment (zk-org/zk#577) <https://github.com/zk-org/zk/pull/577#issuecomment-3356292037>
> On first try I didn't get any diagnostics showing.
> Will try again tomorrow, probably something I've overlooked / done wrong. It's late evening.
> 

During my testing, I sometimes used `./zk edit`, only to realise that this doesn’t change the binary which Vim invokes as an LSP. Maybe you made the same mistake?

> But just to be clear, all that needs to be done is to set a diagnostic level for the `missing-backlink` key, and open a note that has missing backlinks?
> 

Correct. I’ve mostly tried warning.

> Should I expect the diagnostics messages to be displayed inline somewhere? When retrieving with `trouble`, I get a message that no diagnostics available for the 
> 
> 
> Message ID: ***@***.***>
They’re shown inline at the end of the file.
-- 
Hugo

tjex (approved):
Tested all diagnostic positions and levels. All working and tests too. Thanks for the documentation as well.

WhyNotHugo:
Rebased and squashed.

I think the "squash and merge" feature on GitHub discards commit messages. In past PRs I see that the resulting commit message was discarded and the resulting commit has no description, just the PR title.

WhyNotHugo:
> Ahhh, the lines were there, but they don't show if there is no visible empty space after the last line in the buffer window. I needed to create some lines, then delete them, so that I could expose that space

I think this wasn't a problem for me due to `scrolloff=1`, set by vim-sensible.

tjex:
Sounds reasonable, but I have scrollof set to 8.. So that confuses me.

PR #579, opened 2025-09-17
 Format code using gofmt 

WhyNotHugo:
When editing code, newly written code cannot be trivially formatted with
gofmt, since that also changes existing bits of code.

Reformat all code using gofmt, so that future editions can rely on gofmt
without introducing unrelated noise in the diff.


tjex:
Great!

PR #586, opened 2025-09-30
Test for #569, please ignore

WhyNotHugo:


WhyNotHugo:
Looks like CI runs on `refs/remotes/pull/586/merge`, not `refs/remotes/pull/586/head`, so thees results are bogus.

WhyNotHugo:
CI now ran with 000cefb62907292200bea7e4b03cea0afe2193bf, which is identical to f477aacf0dd2fb1812529fc1e54456d36ce5d15e.

Tests for it failed, so the root cause of the failure is not a change in this repository, but something which is fetched non-deterministically.

PR #622, opened 2025-12-03
lsp: implement diagnostic for self-referential links

WhyNotHugo:
A note linking to itself is almost always the result of a mistaken attempting to add a link to a note with a similar name.

Implements: https://github.com/zk-org/zk/discussions/619


tjex (approved):
Working on my end. Updated the default config as well as tesh cases for it. Thank you once again!

PR #631, opened 2025-12-15
docs: mention `zk-emacs` (Emacs integration)

mcookly:
I've been working on an [Emacs package](https://codeberg.org/mcookly/zk-emacs) which integrates `zk` with Emacs over the past six months or so. Currently it's a stable "alpha" project, so I thought it best to weather it with other users.

I'm happy to provide further information as needed!


tjex:
Great! V cool that we're branching into the 'other' world :3

PR #638, opened 2026-01-15
Use AST to find links in LSP

### LINKED ISSUE #630: LSP wrongly parses commented link as "real" links
### Check if applicable

### Describe the bug

The LSP parses commented out links as "real" links, and does not report them as missing backlinks. `zk edit --missing-backlink` properly ignores commented-out links.

### How to reproduce?

Example:

    

### zk configuration

```toml
[lsp.diagnostics]
dead-link = "error"
missing-backlink = { level = "warning", position = "last-section" }
```

### Environment

```bash
zk 0.15.2
```

WhyNotHugo:
Commented links are recognised by the LSP as "real" links, despite them not being actual links on the document. Parsing by the LSP is done with its own bespoke logic, but we already have a perfectly working markdown parser which we use in non-LSP contexts.

Use the same markdown parser in the LSP, so we don't have to manually work around all the edge cases which it already handles. Ensure that the cli and LSP both provide consistent results.

This diff is best reviewed in side-by-side mode; many functions were replaced with new ones, and unified diff mode entangles both versions, yielding unreadable results.

Fixes: https://github.com/zk-org/zk/issues/630

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,9 @@ Format: `<description> (by <contributor>, <pr number>)`
 
 ### Fixed
 
-...
+- Ignore commented links for LSP diagnostics. Use an AST to parse files, fixing
+  other similar edge cases. (by @WhyNotHugo, 638)
+
 
 ## 0.15.2
 
--- a/internal/adapter/lsp/document.go
+++ b/internal/adapter/lsp/document.go
@@ -3,12 +3,16 @@ package lsp
 import (
 	"net/url"
 	"path/filepath"
-	"regexp"
 	"strings"
 	"unicode/utf16"
 
 	"github.com/tliron/glsp"
 	protocol "github.com/tliron/glsp/protocol_3_16"
+	"github.com/yuin/goldmark"
+	"github.com/yuin/goldmark/ast"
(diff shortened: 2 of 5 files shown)


WhyNotHugo:
v2: added changelog entry

tjex (approved):
Beautiful! I didn't get through all the custom markdown and wikilink logic, but enough to be confident. This is more robust than what we have anyway and we can iron out any kinks down the track if they crop up. Thanks for this!

CHANGELOG.md
```diff
@@ -11,7 +11,8 @@ Format: `<description> (by <contributor>, <pr number>)`
 
 ### Fixed
 
-...
+- Ignore commented links for LSP diagnostics. Use an AST to parse files, fixing
+  other similar edge cases.
```
tjex: @WhyNotHugo Your name and the pr number isn't in there. Should we leave them out in general do you think? I kept that pattern from what Michael had set up, and it is potentially useful to track down where something changed. But that is of course possible in many other ways. 

It's also nice for occasional contributors to be recognised in the changelog too.

WhyNotHugo: I didn't have a PR number when I committed this. Updated.

PR #640, opened 2026-01-18
Various linting fixes

WhyNotHugo:
All these tweaks were pointed out by `gopls`, and most of them were auto-fixes done via `gopls`.

They're mostly idiomatic changes, although a few have some performance impact (like using a single `append()` instead of appending inside a loop.


tjex:
I feel cleansed... What command did you use to find all these? I'm only getting warnings about the toml struct entries everywhere...

WhyNotHugo:
I just used `gopls` with staticcheck enabled: https://git.sr.ht/~whynothugo/dotfiles/tree/2a1c65491b4/item/home/.config/nvim/init.lua#L458-464

PR #642, opened 2026-01-19
Replace REGEXP queries with LIKE

WhyNotHugo:
Dramatically improves indexing performance.

See d7ef0e664c7f65d81726a6a4541b2d373df0127b for most details.

PR #643, opened 2026-01-19
Add helper to generate testdata

WhyNotHugo:
For (reproducible) performance benchmarks.

Extracted from https://github.com/zk-org/zk/pull/642


tjex (approved):
Looks good, thank you :D

PR #646, opened 2026-01-19
Resolve links relative to notebook dir

WhyNotHugo:
Reproduces the bug where running the following outside the notebook causes the path to be resolved against the current working directory instead of the notebook directory:

    zk list --notebook-dir=/path/to/notebook --linked-by inbox/note.md

### DIFF
--- a/internal/core/fs_test.go
+++ b/internal/core/fs_test.go
@@ -3,6 +3,7 @@ package core
 import (
 	"path/filepath"
 	"slices"
+	"strings"
 )
 
 // fileStorageMock implements an in-memory FileStorage for testing purposes.
@@ -61,7 +62,11 @@ func (fs *fileStorageMock) DirExists(path string) (bool, error) {
 }
 
 func (fs *fileStorageMock) IsDescendantOf(dir string, path string) (bool, error) {
-	panic("not implemented")
+	rel, err := filepath.Rel(dir, path)
+	if err != nil {
+		return false, err
+	}
+	return !filepath.IsAbs(rel) && !strings.HasPrefix(rel, ".."), nil
 }
 
 func (fs *fileStorageMock) Read(path string) ([]byte, error) {
--- a/internal/core/notebook.go
+++ b/internal/core/notebook.go
@@ -245,7 +245,20 @@ func (n *Notebook) FindCollections(kind CollectionKind, sorters []CollectionSort
 
 // RelPath returns the path relative to the notebook root to the given path.
 func (n *Notebook) RelPath(originalPath string) (string, error) {
-	path, err := n.fs.Abs(originalPath)
+	path := originalPath
+	if !filepath.IsAbs(path) {
+		isInsideNotebook, err := n.fs.IsDescendantOf(n.Path, n.fs.WorkingDir())
+		if err != nil {
+			return path, fmt.Errorf("%v: not a valid notebook path: %w", originalPath, err)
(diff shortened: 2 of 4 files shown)


tjex:
Sorry, had a lot to catch up on here and overlooked that you requested a review! Will look tomorrow. Thanks, as always for the efforts here 🙏

tjex:
@WhyNotHugo I'm getting a relative path to the cwd with a build from dev and your pr.

~Strangely though, the build from dev returns all notes in my notebook, while your pr returns just the links... So something else is at play there.~

But on the note of relative paths to cwd instead of notebook dir. Is this in fact a bug? I myself would expect that the paths _should_ be returned relative to the cwd if I was executing the command outside the notebook. Arguably its a bit redundant as there is the `--format {{path}}` option.

WhyNotHugo:
I think paths should be relative to cwd only if they point to a path inside the notebook. Otherwise, they should be relative to the notebook root.

A relative path that points _outside_ the notebook makes no sense, but it's very convenient to use relative-to-notebook-root from any location.

tjex:
But by definition, paths will always point to a path inside a notebook as zk is only ever indexing notes inside a notebook?... I don't understand how any command would return paths outside of a notebook.

Are you perhaps meaning that paths should only be relative to cwd, if the user is executing the command _inside_ a notebook. While if they execute a command _outside_ the notebook, then the paths should be relative to the notebook root?

WhyNotHugo:
Almost, but with one distinction: 

If cwd is outside the notebook path, but resolving it relative to cwd points to a path inside the notebook, then the path is resolved relative to cwd.

WhyNotHugo:
So my approach is:

- resolve relative to cwd
- does it point inside notebook? Use that
- else: resolve relative to notebook

tjex:
@WhyNotHugo Coming back to this again hit some other problem?

Tried both with HEAD and this PR.

Seem's like I can't point to a note that _isn't_ at the root of the notebook.

```
~/wikis/burn/notebook
$ -> ls foo
x-hh8v.md

~/wikis/burn/notebook
$ -> zk list --linked-by=foo/x-hh8v.md 
zk: error: could not find notes at: foo/x-hh8v.md

~/wikis/burn/notebook
$ -> zk list --notebook-dir=/home/tjex/wikis/burn/notebook --linked-by ./foo/x-hh8v.md
zk: error: could not find notes at: foo/x-hh8v.md

~
$ -> zk list --notebook-dir=/home/tjex/wikis/burn/notebook --linked-by ./foo/x-hh8v.md
zk: error: could not find notes at: foo/x-hh8v.md
```

Here pointing to a file at the root of the notebook.

```
~
$ -> zk list --notebook-dir=/home/tjex/wikis/burn/notebook --linked-by ./2lp4.md
Chicken ... 7 months ago
foo ... 11 months ago
test title ... 3 weeks ago

~/wikis/burn/notebook
$ -> zk list --linked-by ./2lp4.md 
Chicken ... 7 months ago
foo ... 11 months ago
test title ... 3 weeks ago
```

I remember that there was some work being done on the hrefs logic (that was a suboptimal, because the function just returned the first instance of a _filename_ match, even if that file existed in a different directory, mentioned here https://github.com/zk-org/zk/issues/684#issuecomment-4151528520). Not sure if something got merged related to this?

tjex (commented):
See last comment.

PR #658, opened 2026-02-16
Link weave

matze:
This adds links to my [weave](https://github.com/matze/weave) project to the README.md and the static site doc page even though it is actually dynamic.


WhyNotHugo:
Applied, thanks!

PR #664, opened 2026-03-04
Simplify error handling

WhyNotHugo:
Remove the internal `errors` module in favour of directly using the stdlib.

See individual commit messages for clear details.


WhyNotHugo:
A few test fail because they expect the previous less-accurate error messages. I'll update them.

tjex (approved):
Fabulous! Thanks as well for improving the clarity of error output as well, in many many places 🥇 Legend

WhyNotHugo:
> Thanks as well for improving the clarity of error output as well, in many many places

That was mostly a side effect: removing the helper-wrapper meant that I had to copy-paste them, but I might as well write more accurate ones.

It's quite likely that nobody will ever see some of these :joy:

tjex:
Haha. It's true. But good work in the software is invisible :wink:

PR #667, opened 2026-03-07
Prevent out-of-bounds crashes of `textDocument/completion`

virusbb001:
## Description

The `zk` LSP server panics and crashes when receiving a `textDocument/completion` request with a `position.character` that exceeds the current line's length.

## Panic Log

```text
panic: runtime error: slice bounds out of range [:8] with capacity 7

goroutine 13 [running]:
github.com/zk-org/zk/internal/adapter/lsp.(*document).LookBehind(0x8d3a25?, {0x708120?, 0xc0?}, 0x2)
        github.com/zk-org/zk/internal/adapter/lsp/document.go:144 +0x165
github.com/zk-org/zk/internal/adapter/lsp.(*Server).buildInvokedCompletionList(0xc000116600, 0xc0001a7d40, 0xc0004ed590, {0x2c8e10?, 0xc0?})
        github.com/zk-org/zk/internal/adapter/lsp/server.go:647 +0x4b
...
```

## Reproduction Context

1. Run `zk lsp`.
1. Send `initialize` request and wait response.
    * rootUri is `file:///path/to/notebooks/`
1. Send `initialized` notification.
1. Send `textDocument/didOpen` notification.
    * `{"textDocument": {"version": 0, "text": "# Zk\n\n#abc #z\n", "languageId": "markdown", "uri": "file:///path/to/notebooks/Zk.md"}}`
1. Wait `textDocument/publishDiagnostics` notification.
1. Send `textDocument/completion` request.
    * `{"position": {"character": 8, "line": 2}, "context": {"triggerKind": 1}, "textDocument": {"uri": "file:///path/to/notebooks/Zk.md"}}`
1. LSP Server caused panic.

I think this is a client bug (textDocument/completion should be called after sending textDocument/didChange notification). However, I want to avoid causing panic even the LSP client sends an invalid request.

### DIFF
--- a/internal/adapter/lsp/document.go
+++ b/internal/adapter/lsp/document.go
@@ -322,7 +322,8 @@ func (d *document) IsTagPosition(position protocol.Position, noteContentParser c
 		return false
 	}
 	line := lines[lineIdx]
-	if len(line) < charIdx {
+	utf16Len := len(utf16.Encode([]rune(line)))
+	if utf16Len < charIdx {
 		return false
 	}
 	// https://github.com/zk-org/zk/issues/144#issuecomment-1006108485
--- a/internal/adapter/lsp/document.go
+++ b/internal/adapter/lsp/document.go
@@ -318,7 +318,13 @@ func (d *document) IsTagPosition(position protocol.Position, noteContentParser c
 	lines := strutil.CopyList(d.GetLines())
 	lineIdx := int(position.Line)
 	charIdx := int(position.Character)
+	if len(lines) <= lineIdx {
+		return false
+	}
 	line := lines[lineIdx]
+	if len(line) <= charIdx {
+		return false
+	}
 	// https://github.com/zk-org/zk/issues/144#issuecomment-1006108485
 	line = line[:charIdx] + "ZK_PLACEHOLDER" + line[charIdx:]
 	lines[lineIdx] = line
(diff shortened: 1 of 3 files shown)


internal/adapter/lsp/document.go
```diff
@@ -314,7 +318,13 @@ func (d *document) IsTagPosition(position protocol.Position, noteContentParser c
 	lines := strutil.CopyList(d.GetLines())
 	lineIdx := int(position.Line)
 	charIdx := int(position.Character)
+	if len(lines) <= lineIdx {
+		return false
+	}
 	line := lines[lineIdx]
+	if len(line) < charIdx {
+		return false
+	}
```
WhyNotHugo: `charIdx` is clamped to the UTF-16-length of the line, but `len(line)` won't operate on UTF-16.

I think there's cases where invalid input would slip by

WhyNotHugo: `雨果` is 6 bytes in UTF-8 and 2 UTF-16 units.

So value 4 would get accepted here, but is out-of-bounds.

WhyNotHugo (changes requested):

internal/adapter/lsp/document.go
virusbb001: Thanks for the review! I have corrected it, so please review again.

WhyNotHugo (approved):
LGTM

tjex:
Thanks both. Also for the changelog addition :1st_place_medal:

PR #669, opened 2026-03-16
Use standard goldmark link parser for links

### LINKED ISSUE #668: LSP does not return definition when link is broken across newlines
### Check if applicable

- [ ] I'm willing to help fix the problem and contribute a pull request

### Describe the bug

A working link will not return a LSP definition if it crosses newlines, i.e. trying to jump to either
```markdown
[fooo](ab
cd)

[foo
o](abcd)
```
does not work and produces "No results found for `lsp_definitions`".

### How to reproduce?

1. Create a note with the previous content.
2. Go to the line and enter `:lua vim.lsp.buf.definition()`

### zk configuration

```toml
This is independent of the zk configuration.
```

### Neovim configuration

```viml
return {
  {
    "zk-org/zk-nvim",
    config = function()
      require("zk").setup({})
    end
  }
}
```

### Environment

```bash
zk 0.15.2
system: Linux 6.14.0-37-generic x86_64 GNU/Linux
NVIM v0.11.5
Build type: Release
LuaJIT 2.1.1741730670
Run "nvim -V1 -v" for more info
```

WhyNotHugo:
Using our own custom parser to extract the source position of links introduced awkward edge cases.

Use the standard parser, and scan for the position of links afterwards.
We still need to manually scan for the start and end of links, since
goldmark does not expose this. Hopefully, this will change upstream at
some point and we can drop our scanning logic.

Fixes: https://github.com/zk-org/zk/issues/668


WhyNotHugo:
With this change, we should be able to starting converging  LSP and non-LSP code branches to use the same NewParser function.

PR #671, opened 2026-03-17
lsp: Provide completion after [[ on lines with multi-byte characters

virusbb001:
## Description

The `zk` LSP server doesn't provide completion items under specific conditions.

## Reproduction Context

1. Open `tests/fixtures/completion/Item2.md` with the editor that supports LSP.
2. Put the cursor at the end of line 4. (`日本語 [[|`)
3. Invoke `textCompletions/completion` manually.
    * triggerKind is Invoked(1).
4. Check results.
    * Should return all notes as completion items. But returns nothing.


tests/fixtures/completion/.zk/config.toml
```diff
@@ -0,0 +1,202 @@
+# zk configuration file
```
WhyNotHugo: Over 200 lines here is massive, and lots of it are… comments from the original example?

Please reduce this to the strictly minimal necessary lines for tests to work.

internal/adapter/lsp/testdata/completion/Item1.md
```diff
@@ -0,0 +1,4 @@
+# Item1
+
+<!-- Standard completion. Should return all  -->
+[[
```
WhyNotHugo, line 4: Interesting, I didn't know that writing after a pre-existing `[[` could also trigger completion.

On neovim, if I insert after `[[` it works, but if I append, it does not.

Totally unrelated tho, just surprised that this works at all.

internal/adapter/lsp/document_test.go
```diff
@@ -466,3 +466,45 @@ func TestDocument_IsTagPosition(t *testing.T) {
...
+			pos:      protocol.Position{Line: 0, Character: 8},
+			expected: "World",
+		},
+		{
+			name:     "should handle utf-16 correctly",
+			content:  "ビール 焼酎 ジン ワイン ウィスキー",
+			pos:      protocol.Position{Line: 0, Character: 8},
+			expected: "ジン",
+		},
+	}
```
WhyNotHugo, line 504: Can you add an extra test where `Character` is out-of-bounds? Let's try and catch future regressions in this scenario.

internal/adapter/lsp/document.go
```diff
@@ -108,7 +108,11 @@ func (d *document) WordAt(pos protocol.Position) string {
 	if !ok {
 		return ""
 	}
-	return strutil.WordAt(line, int(pos.Character))
+	utf16Bytes := utf16.Encode([]rune(line))
+	charIdx := min(int(pos.Character), len(utf16Bytes))
+	strChar := len(string(utf16.Decode(utf16Bytes[0:charIdx])))
```
WhyNotHugo, line 113: I think we can also get the byte-index of the character at `Position` with something like:

```go
// Return positions of Nth rune:

for i, r := range line {  // i is byte position
    if count == N {
        return i  // byte position
    }
    count++
}
```

This way we don't have to re-encode the entire document as UTF-16.

This might be an unnecessary optimisation at this point tho; feel free to ignore this change if it seems too much; the other are of greater importance.

internal/adapter/lsp/server_test.go
```diff
@@ -0,0 +1,179 @@
...
+func (m *mockFileStorage) Canonical(path string) string {
+	return path
+}
+
+// Get project root directory.
+// It returns absolute path.
+func getProjectRoot() string {
+	_, file, _, _ := runtime.Caller(0)
+	return filepath.Join(filepath.Dir(file), "../../..")
+}
```
WhyNotHugo: Use `Path` from `internal/util/fixtures/fixtures.go` instead of a new function.

internal/adapter/lsp/server_test.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+}
+
+// Read content and create parameter for DidOpen
+func (notebook *notebookFixture) MakeDidOpenParam(noteName string) (protocol.DidOpenTextDocumentParams, error) {
+	path := filepath.Join(notebook.Path, noteName)
+	u := url.URL{
+		Scheme: "file",
+		Path:   filepath.ToSlash(path),
+	}
```
WhyNotHugo: Use `pathToURI` from `internal/adapter/lsp/util.go`.

internal/adapter/lsp/server_test.go
```diff
@@ -0,0 +1,179 @@
...
+
+	t.Run("Return all link completion", func(t *testing.T) {
+		// Position after [[
+		pos := protocol.Position{Line: 3, Character: 2}
+		item, err := server.buildInvokedCompletionList(notebook, doc, pos)
+		assert.Nil(t, err)
+		if len(item) < 1 {
+			t.Error("Number of completion items should be greater than 0")
+		}
+	})
```
WhyNotHugo: Could use https://go.dev/wiki/TableDrivenTests to avoid repetition.

WhyNotHugo (commented):
Thanks, generally looks good.

tests/fixtures/completion/.zk/config.toml
virusbb001: > Over 200 lines here is massive, and lots of it are… comments from the original example?

Yes, I just ran `zk init` in the fixture directory.
I will remove unused comments.

internal/adapter/lsp/document.go
virusbb001: Thanks for the suggestions. But I'll leave it as is for now to keep this PR focused on the other changes.

virusbb001:
@WhyNotHugo Thank you for your review! I have corrected the issues you pointed out, so please review it again.

WhyNotHugo (approved):
LGTM, thanks!

PR #672, opened 2026-03-18
Add tests for bogus handling of UTF-16

WhyNotHugo:
These need review, but I'm pretty sure that the tests are correct and that our logic is broken for these edge cases.


tjex:
Had an initial look. I think I'm following along with the tests correctly... But I'm not so knowledgeable in the underlying issue of byte offsetting.. So it's confidence on shaky legs. Most likely your tests are correct though, and the underlying function implementation is buggy. 

I'm pretty sure this offset solution was added in #317 (or this at least touched the underlying code meaningfully).

PR #674, opened 2026-03-19
Remove hardcoded extension

### LINKED ISSUE #644: FindIdsByHref hardcodes extension
### Check if applicable

### Describe the bug

`FindIdsByHref` hardcodes `.md` extension, but the extension is configurable.

This is often not a problem because later queries match the result, but adds overhead on each search when we'd have an exact match.

### Version

Valid as of b601cdcc4d26907d2f8ce3d2957249ab41ec4bdd

WhyNotHugo:
Fixes: https://github.com/zk-org/zk/issues/644

### DIFF
--- a/internal/adapter/lsp/server_test.go
+++ b/internal/adapter/lsp/server_test.go
@@ -73,8 +73,8 @@ func TestServer_buildInvokedCompletionList(t *testing.T) {
 	}
 	defer db.Close()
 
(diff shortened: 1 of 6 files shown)


WhyNotHugo:
Not entirely convinced about threading the extension all over the place like this, but any alternative approach was far worse.

PR #675, opened 2026-03-22
fix build

virusbb001:
#484 uses `github.com/zk-org/zk/internal/util/errors` but that was removed in #664 .


virusbb001:
It seems I should update tesh tests.

tjex:
...my bad. I fixed some go warnings about the cli cmd struct and that broke the tests. I'm on it :+1: 

And seems the new feature flag wasn't added to the help tesh case in the #484. Added that too

PR #677, opened 2026-03-22
fix(links): unify resolution for Markdown and WikiLinks and fix LSP p…

### LINKED ISSUE #676: Broken Markdown-style link support
### Check if applicable

### Describe the bug

Markdown-style links does not appear to work for `zk list` or for the _LSP_.

Up to now I have used wikilink-style links for internal linking, but wanted to use markdown-style in some instances for readability. However, unless I have missed a config option, they do not seem to work. They do not show up in the `zk list -l` output and the LSP integration in helix returns "not found".

### How to reproduce?

1. Add markdown-style link to existing file.
2. Use the lsp to check the link or run `zk list -l [filename]`

### zk configuration

```toml
# zk configuration file
#
# Uncomment the properties you want to customize.

# NOTE SETTINGS
#
# Defines the default options used when generating new notes.
[note]

# Language used when writing notes.
# This is used to generate slugs or with date formats.
language = "en"

# The default title used for new note, if no `--title` flag is provided.
#default-title = "Untitled"

# Template used to generate a note's filename, without extension.
filename = "{{slug title}}"

# The file extension used for the notes.
extension = "md"

# Template used to generate a note's content.
# If not an absolute path or "~/unix/path", it's relative to .zk/templates/
template = "default.md"

# Path globs ignored while indexing existing notes.
#exclude = [
#    "drafts/*",
#    "log.md"
#]

# Configure random ID generation.

# The charset used for random IDs. You can use:
#   * letters: only letters from a to z.
#   * numbers: 0 to 9
#   * alphanum: letters + numbers
#   * hex: hexadecimal, from a to f and 0 to 9
#   * custom string: will use any character from the provided value
#id-charset = "alphanum"

# Length of the generated IDs.
#id-length = 4

# Letter case for the random IDs, among lower, upper or mixed.
#id-case = "lower"


# EXTRA VARIABLES
#
# A dictionary of variables you can use for any custom values when generating
# new notes. They are accessible in templates with {{extra.<key>}}
[extra]

#key = "value"


# GROUP OVERRIDES
#
# You can override global settings from [note] and [extra] for a particular
# group of notes by declaring a [group."<name>"] section.
#
# Specify the list of directories which will automatically belong to the group
# with the optional `paths` property.
#
# Omitting `paths` is equivalent to providing a single path equal to the name of
# the group. This can be useful to quickly declare a group by the name of the
# directory it applies to.

#[group."<NAME>"]
#paths = ["<DIR1>", "<DIR2>"]
#[group."<NAME>".note]
#filename = "{{format-date now}}"
#[group."<NAME>".extra]
#key = "value"


[group."sources"]
paths = ["sources"]
[group."sources".note]
filename = "@{{extra.author_last}}-{{extra.year_filename}}-{{extra.slug_title}}"


[group."sources/authors"]
paths = ["sources/authors"]
[group."sources/authors".note]
filename = "{{extra.author}}"

[group."main"]
paths = ["main"]
template="main.md"
[group."main".note]
filename = "{{slug title}}"

# MARKDOWN SETTINGS
[format.markdown]

# Format used to generate links between notes.
# Either "wiki", "markdown" or a custom template. Default is "markdown".

link-format = "markdown" # I have also tried wiki, but the issue persists

# Indicates whether a link's path will be percent-encoded.
# Defaults to true for "markdown" format and false for "wiki" format.
link-encode-path = false
# Indicates whether a link's path file extension will be removed.
# Defaults to true.
link-drop-extension = false

# Enable support for #hashtags.
hashtags = true
# Enable support for :colon:separated:tags:.
colon-tags = false
# Enable support for Bear's #multi-word tags#
# Hashtags must be enabled for multi-word tags to work.
multiword-tags = false


# EXTERNAL TOOLS
[tool]

# Default editor used to open notes. When not set, the EDITOR or VISUAL
# environment variables are used.
#editor = "vim"

# Pager used to scroll through long output. If you want to disable paging
# altogether, set it to an empty string "".
#pager = "less -FIRX"

# Command used to preview a note during interactive fzf mode.
# Set it to an empty string "" to disable preview.

# bat is a great tool to render Markdown document with syntax highlighting.
#https://github.com/sharkdp/bat
#fzf-preview = "bat -p --color always {-1}"


# LSP
#
#   Configure basic editor integration for LSP-compatible editors.
#   See https://zk-org.github.io/zk/tips/editors-integration.html
#
[lsp]

[lsp.diagnostics]
# Each diagnostic can have for value: none, hint, info, warning, error

# Report titles of wiki-links as hints.
#wiki-title = "hint"
# Warn for dead links between notes.
dead-link = "error"
# Warn when a note links to itself.
#self-link = "warning"
# Report missing backlinks
#missing-backlink = { level = "hint", position = "bottom" }


[lsp.completion]
# Customize the completion pop-up of your LSP client.

# Show the note title in the completion pop-up, or fallback on its path if empty.
#note-label = "{{title-or-path}}"
# Filter out the completion pop-up using the note title or its path.
#note-filter-text = "{{title}} {{path}}"
# Show the note filename without extension as detail.
#note-detail = "{{filename-stem}}"


# NAMED FILTERS
#
#    A named filter is a set of note filtering options used frequently together.
#
[filter]

# Matches the notes created the last two weeks. For example:
#    $ zk list recents --limit 15
#    $ zk edit recents --interactive
#recents = "--sort created- --created-after 'last two weeks'"


# COMMAND ALIASES
#
#   Aliases are user commands called with `zk <alias> [<flags>] [<args>]`.
#
#   The alias will be executed with `$SHELL -c`, please refer to your shell's
#   man page to see the available syntax. In most shells:
#     * $@ can be used to expand all the provided flags and arguments
#     * you can pipe commands together with the usual | character
#
[alias]
# Here are a few aliases to get you started.

# Shortcut to a command.
ls = "zk list $@"

# Default flags for an existing command.
#list = "zk list --quiet $@"

# Edit the last modified note.
editlast = "zk edit --limit 1 --sort modified- $@"

# Edit the notes selected interactively among the notes created the last two weeks.
# This alias doesn't take any argument, so we don't use $@.
recent = "zk edit --sort created- --created-after 'last two weeks' --interactive"

# Print paths separated with colons for the notes found with the given
# arguments. This can be useful to expand a complex search query into a flag
# taking only paths. For example:
#   zk list --link-to "`zk path -m potatoe`"
#path = "zk list --quiet --format {{path}} --delimiter , $@"

# Show a random note.
#lucky = "zk list --quiet --format full --sort random --limit 1"

# Returns the Git history for the notes found with the given arguments.
# Note the use of a pipe and the location of $@.
#hist = "zk list --format path --delimiter0 --quiet $@ | xargs -t -0 git log --patch --"

# Edit this configuration file.
conf = '$EDITOR "$ZK_NOTEBOOK_DIR/.zk/config.toml"'

e = "zk edit -s 'created' -i"




lit = "zk new --template literature.md --notebook-dir sources"

nlit = "python ./scripts/nlit.py"
nauthor = "python ./scripts/nauthor.py"
nlec = "python ./scripts/nlecture.py"
ncirc = "python ./scripts/ncirc.py"
nmain ="zk new --template main.md --title $@ main"
a = "python3 ./scripts/zllm.py $@"
process_handnotes = "python ./scripts/process_handnotes.py"


wc = "zk list --format '{{word-count}}\t{{title}}' --sort word-count $@"


log = "zk list --quiet --format path --delimiter0 $@ | xargs -0 git log --patch --"
```

### Environment

```bash
zk 0.15.2
system: Linux 6.19.7-200.fc43.x86_64 x86_64 GNU/Linux
```

sfgartland:
…anic

Markdown and WikiLinks are now resolved using an identical three-step hierarchy:
1. Exact match relative to the vault root.
2. Match relative to the current file.
3. Fallback to partial path matching.

This fixes the issue where Markdown links requiring extensions were not properly indexed, and ensures consistent link resolution across CLI and LSP functionalities.

Additionally, added a nil-pointer check to the LSP definition handler to prevent crashes when resolving definitions for editors that do not provide full TextDocument LinkSupport capabilities.

...

 Fixes #676

### DIFF
--- a/internal/adapter/lsp/server.go
+++ b/internal/adapter/lsp/server.go
@@ -325,7 +325,12 @@ func NewServer(opts ServerOpts) *Server {
 			return nil, err
 		}
 
-		if isTrue(clientCapabilities.TextDocument.Definition.LinkSupport) {
+		linkSupport := false
+		if clientCapabilities.TextDocument != nil && clientCapabilities.TextDocument.Definition != nil {
+			linkSupport = isTrue(clientCapabilities.TextDocument.Definition.LinkSupport)
+		}
+
+		if linkSupport {
 			return protocol.LocationLink{
 				OriginSelectionRange: &link.Range,
 				TargetURI:            target.URI,
@@ -531,13 +536,25 @@ func (s *Server) notebookOf(doc *document) (*core.Notebook, error) {
 //  2. Find any occurrence of the href in a note path (substring)
 //  3. Match the href as a term in the note titles
 func (s *Server) noteForLink(link documentLink, notebook *core.Notebook) (*Note, error) {
-	note, err := s.noteForHref(link.Href, link.RelativeToDir, notebook)
-	if note == nil && err == nil && link.IsWikiLink {
-		// Try to find a partial href match.
-		note, err = notebook.FindByHref(link.Href, true)
+	if strutil.IsURL(link.Href) {
+		return nil, nil
 	}
-	if note == nil || err != nil {
-		return nil, err
+
+	// 1. Try relative to the current file (relativeToDir)
+	note, _ := s.noteForHref(link.Href, link.RelativeToDir, notebook)
+
+	// 2. Try relative to vault root
+	if note == nil {
+		note, _ = notebook.FindByHref(link.Href, false)
+	}
+
+	// 3. Fallback to partial matching for both link types
+	if note == nil {
+		note, _ = notebook.FindByHref(link.Href, true)
+	}
+
+	if note == nil {
+		return nil, fmt.Errorf("failed to resolve href: %s", link.Href)
 	}
 
 	joinedPath := filepath.Join(notebook.Path, note.Path)
--- a/internal/adapter/sqlite/note_dao.go
+++ b/internal/adapter/sqlite/note_dao.go
@@ -296,31 +296,48 @@ func (d *NoteDAO) findIDsByHrefs(hrefs []string, allowPartialHrefs bool) ([]core
 // FindIdsByHref finds note IDs which match the given href string.
 // This implements logic similar to NoteIndex.linkMatchesPath.
 func (d *NoteDAO) FindIdsByHref(href string, allowPartialHref bool) ([]core.NoteID, error) {
+	var err error
 	// Remove any anchor at the end of the HREF, since it's most likely
 	// matching a sub-section in the note.
 	href = strings.SplitN(href, "#", 2)[0]
 
 	href = strings.NewReplacer("%", "\\%", "_", "\\_").Replace(href)
+	
+	// Treat leading slash as relative to the notebook root.
+	rootHref := strings.TrimPrefix(href, "/")
 
-	// Prioritise exact match with extension.
-	id, err := d.FindIDByPath(href + "." + d.extension)
-	if err != nil {
-		return nil, err
-	}
-	if id.IsValid() {
-		return []core.NoteID{id}, nil
+	// 1. Prioritize exact match (either literal href or root-relative).
+	for _, h := range []string{href, rootHref} {
+		id, err := d.FindIDByPath(h)
+		if err != nil {
+			return nil, err
+		}
+		if id.IsValid() {
+			return []core.NoteID{id}, nil
+		}
+
+		// Try with extension.
+		if !strings.HasSuffix(h, "."+d.extension) {
+			id, err = d.FindIDByPath(h + "." + d.extension)
+			if err != nil {
+				return nil, err
+			}
+			if id.IsValid() {
+				return []core.NoteID{id}, nil
+			}
+		}
 	}
 
 	var ids []core.NoteID
 	if allowPartialHref {
 		// Filename (not path) contains 'href' anywhere.
-		ids, err = d.findIDsByFilenameLike("%" + href + "%")
+		ids, err = d.findIDsByFilenameLike("%" + rootHref + "%")
 		if len(ids) > 0 || err != nil {
 			return ids, err
 		}
 
 		// Path contains 'href' anywhere.
-		ids, err = d.findIDsByPathLike("%" + href + "%")
+		ids, err = d.findIDsByPathLike("%" + rootHref + "%")
 		if len(ids) > 0 || err != nil {
 			return ids, err
 		}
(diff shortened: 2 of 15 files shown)


WhyNotHugo:
The correct order is:

- Match relative to the current file.
- Exact match relative to the vault root.
- Fallback to partial path matching.

If we always default to the root, then `x/a.md` becomes unreachable if `a.md` exists, and links would change their destination when creating the latter.

In general terms, markdown links are relative to the current file. The other two behaviours are fallbacks for unusual usages.

sfgartland:
> If we always default to the root, then `x/a.md` becomes unreachable if `a.md` exists, and links would change their destination when creating the latter.

I had not considered this issue. I have always defaulted to using the zettel root as a quasi-absolute path for the sake of making refractoring links on move easier, but I can see the other use-cases as well.

I've amended my commit to reflect the followin priority:
> 1. Match relative to the current file.
> 2. Exact match relative to the vault root.
> 3. Fallback to partial path matching.

And also added relevant end-to-end and unit tests

internal/adapter/lsp/server.go
```diff
@@ -325,7 +325,12 @@ func NewServer(opts ServerOpts) *Server {
 			return nil, err
 		}
 
-		if isTrue(clientCapabilities.TextDocument.Definition.LinkSupport) {
+		linkSupport := false
+		if clientCapabilities.TextDocument != nil && clientCapabilities.TextDocument.Definition != nil {
+			linkSupport = isTrue(clientCapabilities.TextDocument.Definition.LinkSupport)
+		}
+
+		if linkSupport {
```
WhyNotHugo, line 333: Why is this change required here?

internal/adapter/sqlite/note_dao.go
```diff
@@ -296,31 +296,48 @@ func (d *NoteDAO) findIDsByHrefs(hrefs []string, allowPartialHrefs bool) ([]core
...
-		return []core.NoteID{id}, nil
+	// 1. Prioritize exact match (either literal href or root-relative).
+	for _, h := range []string{href, rootHref} {
+		id, err := d.FindIDByPath(h)
+		if err != nil {
+			return nil, err
+		}
+		if id.IsValid() {
+			return []core.NoteID{id}, nil
+		}
```
WhyNotHugo, line 317: In `x/a.md`, a link `b.md` will prefer note `b.md`, but should prefer `x/b.md`.

PR #679, opened 2026-03-23
Slightly simplify link resolution

WhyNotHugo:
Mostly inline trivial functions and document another one.

There's more to be done in this area, but I'll make progressive refinements here.


WhyNotHugo:
Something that keeps confusing me is `FindIDByHref`, which returns the first result, but with no clear criteria. If both `a/x.md` and `b/x.md` exist, the return value is not deterministic.

PR #690, opened 2026-04-06
Pass shell explicitly to functions that need it

### LINKED ISSUE #689: config.Tool.Shell in notebook config doesn't apply
### Check if applicable

### Describe the bug

It looks like the value in the configuration for `tool.shell` doesn't actually get used if I understand the code correctly. Running commands uses the environment variable pulled from [exec.CommandFromString](https://github.com/zk-org/zk/blob/dev/internal/util/exec/exec_unix.go#L12-L20), which if not already set is set to the value of the config value in [container.NewContainer](https://github.com/zk-org/zk/blob/dev/internal/cli/container.go#L87-L890). However it seems (at least in my case) the configuration isn't read until after that. I think it's loaded as part of [NotebookStore.Open](https://github.com/zk-org/zk/blob/dev/internal/core/notebook_store.go#L63-L66) which happens after the container is created (and the environment variable set).

So maybe having another check to set the environment variable after loading the notebook configuration. At least adding this check to after loading the configuration in the `NotebookStore.Open` solves the issue for me (it means that global config would take priority since that sets it earlier)
```go
	// Set the default shell if not already set
	if osutil.GetOptEnv("ZK_SHELL").IsNull() && !config.Tool.Shell.IsEmpty() {
		os.Setenv("ZK_SHELL", config.Tool.Shell.Unwrap())
	}
```

### How to reproduce?

1. Use `fish` shell
2. Try to use an alias with non-fish syntax
3. Run `zk list`
```
fish: $@ is not supported. In fish, please use $argv.
cd "/home/arden/Zettle" && zk list --quiet $@
```

### zk configuration

```toml
[tool]
shell = "/bin/bash"

[alias]
list = "zk list --quiet $@"
```

### Environment

```bash
zk 0.15.2
system: Linux 6.19.10-arch1-1 x86_64 GNU/Linux
```

WhyNotHugo:
The shell specified in the configuration is passed internally by setting the ZK_SHELL environment variable in one place, and then assuming it is set in functions that need it.

This makes reasoning about ordering nearly impossible, since there's no variable being passed, but a global variable being set instead.

Pass the shell explicitly to sites that need to use it.

Fixes: https://github.com/zk-org/zk/issues/689

### DIFF
--- a/internal/adapter/editor/editor.go
+++ b/internal/adapter/editor/editor.go
@@ -15,11 +15,12 @@ import (
 // Editor represents an external editor able to edit the notes.
 type Editor struct {
 	editor string
+	shell  string
 }
 
 // NewEditor creates a new Editor from the given editor user setting or the
 // matching environment variables.
-func NewEditor(editor opt.String) (*Editor, error) {
+func NewEditor(editor opt.String, shell string) (*Editor, error) {
 	editor = osutil.GetOptEnv("ZK_EDITOR").
 		Or(editor).
 		Or(osutil.GetOptEnv("VISUAL")).
@@ -29,7 +30,7 @@ func NewEditor(editor opt.String) (*Editor, error) {
 		return nil, fmt.Errorf("no editor set in config")
 	}
 
-	return &Editor{editor.Unwrap()}, nil
+	return &Editor{editor: editor.Unwrap(), shell: shell}, nil
 }
 
 // Open launches the editor with the notes at given paths.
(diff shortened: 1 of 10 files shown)


Nedra1998:
Thanks for the quick work! This does fix the issue I was having (tested on my system), and seems much easier for reasoning about the order of when things are loaded and used.

WhyNotHugo:
Thanks for confirming. Yeah, the main issue was that we could not reason about the order of events, and *apparently* they were wrong.

tjex:
Seconded on the fast turn around on this, and everything else too. Total fire!!

PR #701, opened 2026-04-29
Make tag positions part of the index

ghost:
This is related to what was discussed in #699.

Please let me know if these changes are acceptable.

### DIFF
--- a/internal/adapter/markdown/markdown.go
+++ b/internal/adapter/markdown/markdown.go
@@ -2,9 +2,8 @@ package markdown
 
 import (
 	"bufio"
-	"fmt"
+
 	"net/url"
-	"regexp"
 	"strings"
 
 	"github.com/mvdan/xurls"
@@ -19,7 +18,7 @@ import (
 	"github.com/zk-org/zk/internal/util"
 	"github.com/zk-org/zk/internal/util/opt"
 	strutil "github.com/zk-org/zk/internal/util/strings"
-	"github.com/zk-org/zk/internal/util/yaml"
+	yaml "github.com/zk-org/zk/internal/util/yaml"
 )
 
 // Parser parses the content of Markdown notes.
@@ -36,6 +35,8 @@ type ParserOpts struct {
 	// Indicates whether :colon:tags: are parsed.
 	ColontagEnabled bool
 }
+
+type FrontMatterExtended = yaml.FrontmatterExtended
 
 // NewParser creates a new Markdown Parser.
 func NewParser(options ParserOpts, logger util.Logger) *Parser {
@@ -65,6 +66,24 @@ func NewParser(options ParserOpts, logger util.Logger) *Parser {
 	}
 }
 
+func tagNames(parsedTags []core.Tag) []string {
+	if parsedTags == nil {
+		return []string{}
+	}
+	check := make(map[string]struct{})
+	res := make([]string, 0)
+
+	for _, tag := range parsedTags {
+		tagName := tag.Name
+		if _, ok := check[tagName]; ok {
+			continue
+		}
+		check[tagName] = struct{}{}
+		res = append(res, tagName)
+	}
+	return res
+}
+
 // ParseNoteContent implements core.NoteContentParser.
 func (p *Parser) ParseNoteContent(content string) (*core.NoteContent, error) {
 	bytes := []byte(content)
@@ -79,8 +98,7 @@ func (p *Parser) ParseNoteContent(content string) (*core.NoteContent, error) {
 	if err != nil {
 		return nil, err
 	}
-
-	frontmatter, err := parseFrontmatter(context, bytes)
+	frontmatter, err := yaml.ParseFrontmatter(bytes)
 	if err != nil {
 		return nil, err
 	}
@@ -89,27 +107,33 @@ func (p *Parser) ParseNoteContent(content string) (*core.NoteContent, error) {
 	if err != nil {
 		return nil, err
 	}
+
 	body := parseBody(bodyStart, bytes)
 
-	tags, err := parseTags(frontmatter, root)
+	tags, err := parseTags(root)
 	if err != nil {
 		return nil, err
 	}
 
+	combinedTags := append(frontmatter.Tags, tags...)
+
 	return &core.NoteContent{
-		Title:    title,
-		Body:     body,
-		Lead:     parseLead(body),
-		Links:    links,
-		Tags:     tags,
-		Metadata: frontmatter.values,
+		Title:        title,
+		Body:         body,
+		Lead:         parseLead(body),
+		Links:        links,
+		Tags:         tagNames(combinedTags),
+		Metadata:     frontmatter.Values,
+		ExtendedTags: combinedTags,
 	}, nil
 }
 
 // parseTitle extracts the note title with its node.
-func parseTitle(frontmatter frontmatter, root ast.Node, source []byte) (title opt.String, bodyStart int, err error) {
-	if title = frontmatter.getString("title", "Title"); !title.IsNull() {
-		bodyStart = frontmatter.end
+func parseTitle(frontmatter *FrontMatterExtended, root ast.Node, source []byte) (title opt.String, bodyStart int, err error) {
+	titleString := frontmatter.GetTitleString()
+	if !titleString.IsNull() {
+		title = titleString
+		bodyStart = frontmatter.End
 		return
 	}
 
@@ -163,51 +187,30 @@ func parseLead(body opt.String) opt.String {
 	return opt.NewNotEmptyString(strings.TrimSpace(lead.String()))
 }
 
-// parseTags extracts tags as #hashtags, :colon:tags: or from the YAML frontmatter.
-func parseTags(frontmatter frontmatter, root ast.Node) ([]string, error) {
-	tags := make([]string, 0)
-
-	// Parse from YAML frontmatter, either:
-	// * a list of strings
-	// * a single space-separated string
-	findFMTags := func(key string) []string {
-		if tags, ok := frontmatter.getStrings(key); ok {
-			return tags
-
-		} else if tags := frontmatter.getString(key); !tags.IsNull() {
-			// Parse a space-separated string list
-			res := []string{}
-			for s := range strings.FieldsSeq(tags.Unwrap()) {
-				s = strings.TrimSpace(s)
-				if len(s) > 0 {
-					res = append(res, s)
-				}
-			}
-			return res
-
-		} else {
-			return []string{}
-		}
+func resolveTags(tags []string, pos int) []core.Tag {
+	// if the tags slice has length larger than one, we assume these are colon separated tags,
+	// that the parser has returned only the start position of the first of them, and we make position adjustments.
+	result := make([]core.Tag, 0)
+	currentPosition := 0
+	for _, t := range tags {
+		result = append(result, core.Tag{Name: t, Pos: pos + currentPosition})
+		currentPosition += len(t) + 1
 	}
+	return result
+}
 
-	for _, key := range []string{"tag", "tags", "keyword", "keywords"} {
-		for _, t := range findFMTags(key) {
-			// Trims any # prefix to support hashtags embedded in YAML
-			// frontmatter, as in Simple Markdown Zettelkasten:
-			// http://evantravers.com/articles/2020/11/23/zettelkasten-updates/
-			tags = append(tags, strings.TrimPrefix(t, "#"))
-		}
-	}
-
+func parseTags(root ast.Node) ([]core.Tag, error) {
 	// Parse #hashtags and :colon:tags:
+	tags := []core.Tag{}
 	err := ast.Walk(root, func(n ast.Node, entering bool) (ast.WalkStatus, error) {
 		if tagsNode, ok := n.(*extensions.Tags); ok && entering {
-			tags = append(tags, tagsNode.Tags...)
+			tagNodePosition := tagsNode.Pos()
+			tagNodeContent := tagsNode.Tags
+			tags = append(tags, resolveTags(tagNodeContent, tagNodePosition)...)
 		}
 		return ast.WalkContinue, nil
 	})
-
-	return strutil.RemoveDuplicates(tags), err
+	return tags, err
 }
 
 // parseLinks extracts outbound links from the note.
@@ -290,84 +293,4 @@ func extractLines(n ast.Node, source []byte) (content string, start, end int) {
 	}
 
 	return
-}
-
-// frontmatter contains metadata parsed from a YAML frontmatter.
-type frontmatter struct {
-	values map[string]any
-	start  int
-	end    int
-}
-
-var frontmatterRegex = regexp.MustCompile(`(?ms)^\s*-+\s*$.*?^\s*-+\s*$`)
-
-func parseFrontmatter(context parser.Context, source []byte) (frontmatter, error) {
-	var front frontmatter
-	front.values = map[string]any{}
-
-	index := frontmatterRegex.FindIndex(source)
-	if index == nil {
-		return front, nil
-	}
-
-	front.start = index[0]
-	front.end = index[1]
-
-	values, err := meta.TryGet(context)
-	if err != nil {
-		return front, err
-	}
-
-	// The YAML parser parses nested maps as map[any]any
-	// instead of map[string]any, which doesn't work with the JSON
-	// marshaller.
-	values = yaml.ConvertMapToJSONCompatible(values)
-
-	// Convert keys to lowercase, because we don't want to be case sensitive.
-	for k, v := range values {
-		front.values[strings.ToLower(k)] = v
-	}
-
-	return front, nil
-}
-
-// getString returns the first string value found for any of the given keys.
-func (m frontmatter) getString(keys ...string) opt.String {
-	if m.values == nil {
-		return opt.NullString
-	}
-
-	for _, key := range keys {
-		key = strings.ToLower(key)
-		if val, ok := m.values[key]; ok {
-			if val, ok := val.(string); ok {
-				return opt.NewNotEmptyString(val)
-			}
-		}
-	}
-	return opt.NullString
-}
-
-// getStrings returns the first string list found for any of the given keys.
-func (m frontmatter) getStrings(keys ...string) ([]string, bool) {
-	if m.values == nil {
-		return nil, false
-	}
-
-	for _, key := range keys {
-		key = strings.ToLower(key)
-		if val, ok := m.values[key]; ok {
-			if val, ok := val.([]any); ok {
-				strs := []string{}
-				for _, v := range val {
-					s := strings.TrimSpace(fmt.Sprint(v))
-					if len(s) > 0 {
-						strs = append(strs, s)
-					}
-				}
-				return strs, true
-			}
-		}
-	}
-	return nil, false
 }
--- a/internal/adapter/sqlite/note_index.go
+++ b/internal/adapter/sqlite/note_index.go
@@ -27,6 +27,7 @@ type dao struct {
 	notes       *NoteDAO
 	links       *LinkDAO
 	collections *CollectionDAO
+	tags        *TaggedNoteDAO
 	metadata    *MetadataDAO
 }
 
@@ -85,6 +86,22 @@ func (ni *NoteIndex) FindLinksBetweenNotes(ids []core.NoteID) (links []core.Reso
 	return
 }
 
+func (ni *NoteIndex) FindTaggedNotes(tagName string, sorters []core.TaggedNoteSorter) (tag []core.TaggedNoteDetail, err error) {
+	err = ni.commit(func(dao *dao) error {
+		tag, err = dao.tags.FindTaggedNotes(tagName, sorters)
+		return err
+	})
+	return
+}
+
+func (ni *NoteIndex) FindTaggedNotesAll(sorters []core.TaggedNoteSorter) (tags []core.TaggedNoteDetail, err error) {
+	err = ni.commit(func(dao *dao) error {
+		tags, err = dao.tags.FindTaggedNotesAll(sorters)
+		return err
+	})
+	return
+}
+
 // FindCollections implements core.NoteIndex.
 func (ni *NoteIndex) FindCollections(kind core.CollectionKind, sorters []core.CollectionSorter) (collections []core.Collection, err error) {
 	err = ni.commit(func(dao *dao) error {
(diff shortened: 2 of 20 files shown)


internal/adapter/markdown/markdown.go
```diff
@@ -89,27 +108,33 @@ func (p *Parser) ParseNoteContent(content string) (*core.NoteContent, error) {
...
 
 // parseTitle extracts the note title with its node.
-func parseTitle(frontmatter frontmatter, root ast.Node, source []byte) (title opt.String, bodyStart int, err error) {
-	if title = frontmatter.getString("title", "Title"); !title.IsNull() {
-		bodyStart = frontmatter.end
+func parseTitle(frontmatter *FrontMatterExtended, root ast.Node, source []byte) (title opt.String, bodyStart int, err error) {
+	titleString, found := frontmatter.GetTitleString()
+	title = opt.NewNotEmptyString(titleString)
+	if found && !title.IsNull() {
+		bodyStart = frontmatter.End
```
WhyNotHugo, line 136: What's the effect of this change?

internal/util/yaml/yaml.go
```diff
@@ -1,56 +1,224 @@
...
+}
 
-	case []any:
-		for i, v2 := range x {
-			x[i] = ConvertToJSONCompatible(v2)
+func trimTag(tag string) string {
+	trimmedTag := strings.TrimPrefix(tag, `"`)
+	trimmedTag = strings.TrimSuffix(tag, `"`)
+	trimmedTag = strings.TrimSpace(tag)
+	trimmedTag = strings.TrimPrefix(trimmedTag, "#")
```
WhyNotHugo: These should operate on `trimmedTag`, not `tag`.

internal/adapter/sqlite/note_index.go
```diff
@@ -247,18 +264,41 @@ func (ni *NoteIndex) Update(note core.Note) error {
...
 }
 
-func (ni *NoteIndex) associateTags(collections *CollectionDAO, noteID core.NoteID, tags []string) error {
+func (ni *NoteIndex) associateTags(dao *dao, note *core.Note, noteID core.NoteID) error {
+	tagIDs, error := ni.associateTagCollections(dao, note, noteID)
+	if error != nil {
+		return error
+	}
+	return ni.associateTagOccurrences(dao, note, noteID, tagIDs)
+}
```
WhyNotHugo, line 282: I believe old tags are never removed?

internal/adapter/sqlite/note_index.go
```diff
@@ -247,18 +264,41 @@ func (ni *NoteIndex) Update(note core.Note) error {
 	return nil
 }
 
-func (ni *NoteIndex) associateTags(collections *CollectionDAO, noteID core.NoteID, tags []string) error {
+func (ni *NoteIndex) associateTags(dao *dao, note *core.Note, noteID core.NoteID) error {
+	tagIDs, error := ni.associateTagCollections(dao, note, noteID)
```
WhyNotHugo: Nitpick: Should use `err`, not `error`.

internal/core/tag.go
```diff
@@ -0,0 +1,76 @@
...
+type TagSortField int
+
+const (
+	// Sort by the tag name.
+	TagSortName TagSortField = 1 << iota
+	// Sort by the title of the note.
+	TagSortTitle
+)
+
+// CollectionSortersFromStrings returns a list of CollectionSorter from their string
```
WhyNotHugo:
```suggestion
// TagSortersFromString returns a list of TagSorter from their string
```

internal/core/tag.go
```diff
@@ -0,0 +1,76 @@
...
+	Positions []int  `json:"pos"`
+}
+
+// CollectionSorter represents an order term used to sort a list of collections.
+type TagSorter struct {
+	Field     TagSortField
+	Ascending bool
+}
+
+// CollectionSortField represents a collection field used to sort a list of collections.
```
WhyNotHugo:
```suggestion
// CollectionSortField represents a tag field used to sort a list of tags.
```

internal/core/tag.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+type TagDetail struct {
+	Tag       string `json:"tag"`
+	Title     string `json:"title"`
+	Path      string `json:"path"`
+	Positions []int  `json:"pos"`
+}
+
+// CollectionSorter represents an order term used to sort a list of collections.
```
WhyNotHugo:
```suggestion
// TagSorter struct represents an order term used to sort a list of tags.
```

internal/core/tag.go
```diff
@@ -0,0 +1,76 @@
...
+		sorter, err := TagSorterFromString(strs[i])
+		if err != nil {
+			return sorters, err
+		}
+		sorters = append(sorters, sorter)
+	}
+	return sorters, nil
+}
+
+// CollectionSorterFromString returns a CollectionSorter from its string representation.
```
WhyNotHugo:
```suggestion
// TagSorterFromString returns a TagSorter from its string representation.
```

internal/util/misc/misc.go
```diff
@@ -0,0 +1,20 @@
+package misc
+
+func RemoveDuplicates[T comparable](input []T) []T {
```
WhyNotHugo: I think you could just use the existing `RemoveDuplicates` from internal/util/strings/strings.go and avoid this new function.

internal/util/yaml/yaml.go
```diff
@@ -1,56 +1,224 @@
...
+	err := yml.Unmarshal(frontMatterBytes, &rootNode)
+
+	if err != nil {
+		return nil, err
+	}
+
+	frontmatter.Root = &rootNode
+	frontmatter.Bytes = frontMatterBytes
+
+	if !(len(rootNode.Content) > 0) {
```
WhyNotHugo:
```suggestion
	if (len(rootNode.Content) <= 0) {
```

internal/util/yaml/yaml.go
```diff
@@ -1,56 +1,224 @@
...
+		return &frontmatter, nil
+	}
+
+	mapping := rootNode.Content[0]
+	if !(mapping.Kind == yml.MappingNode) {
```
WhyNotHugo:
```suggestion
	if (mapping.Kind != yml.MappingNode) {
```

WhyNotHugo (commented):
I've only gone through half of this, it's a bit long. So far looks generally good, this is some preliminary feedback.

ghost:
I've just noticed that removing the generic `RemoveDuplicates` will cause the association to break if we have duplicates in colon separated tags, because they are resolved to the same position.

It can either be put back, or the parser modified to resolve them to different positions.

ghost:
It turns out that it's only a small change to correct this.

internal/adapter/markdown/markdown.go
WhyNotHugo: The refactor in this function is unnecessary, only a single function's name has changed.

internal/core/tag.go
WhyNotHugo, line 1: This module really deals with `TaggedNote` rather than `Note`; the entity it's representing is a note with a tag, not a tag itself.

Also `TagSorter` -> `TaggedNoteSorter`. Again; it's referring to fields in the note, note the tag. Same applies for `internal/core/tag_format.go`.

I mostly want to keep things intuitive to grasp for future readers (including ourselves in 9 months).

(logic itself is all sound)

**Edit**: could also be called `TagAssociation` I guess?

internal/adapter/sqlite/note_index.go
```diff
@@ -85,6 +86,22 @@ func (ni *NoteIndex) FindLinksBetweenNotes(ids []core.NoteID) (links []core.Reso
 	return
 }
 
+func (ni *NoteIndex) FindTag(tagName string, sorters []core.TagSorter) (tag []core.TagDetail, err error) {
+	err = ni.commit(func(dao *dao) error {
+		tag, err = dao.tags.FindTag(tagName, sorters)
+		return err
+	})
```
WhyNotHugo, line 93: Since you're already assigning to the return values, the intermediate assignment for `err` is unnecessary.

```suggestion
	ni.commit(func(dao *dao) {
		tag, err = dao.tags.FindTag(tagName, sorters)
	})
```

internal/adapter/sqlite/note_index.go
```diff
@@ -85,6 +86,22 @@ func (ni *NoteIndex) FindLinksBetweenNotes(ids []core.NoteID) (links []core.Reso
...
+	return
+}
+
+func (ni *NoteIndex) FindAllTags(sorters []core.TagSorter) (tags []core.TagDetail, err error) {
+	err = ni.commit(func(dao *dao) error {
+		tags, err = dao.tags.FindAllTags(sorters)
+		return err
+	})
+	return
+}
```
WhyNotHugo: Same as above:

```suggestion
func (ni *NoteIndex) FindAllTags(sorters []core.TagSorter) (tags []core.TagDetail, err error) {
	ni.commit(func(dao *dao) {
		tags, err = dao.tags.FindAllTags(sorters)
	})
	return
}
```

internal/adapter/sqlite/note_index.go
```diff
@@ -247,18 +264,50 @@ func (ni *NoteIndex) Update(note core.Note) error {
...
 		}
+		tagIDs[tag] = tagID
 		_, err = collections.Associate(noteID, tagID)
 		if err != nil {
-			return err
+			return nil, err
 		}
 	}
+	return tagIDs, nil
+}
```
WhyNotHugo, line 300: This is not blocking for the PR and we can do this separately later:

Adding a tag here doesn't need two separate queries, we can `INSERT` with a `SELECT` in a subquery. This reduces a whole query (which in turn is a cgo function call, plus the query itself).

internal/core/note.go
```diff
@@ -47,6 +52,8 @@ type Note struct {
 	Links []Link
 	// List of tags found in the content.
 	Tags []string
```
WhyNotHugo, line 54: Can we deprecate this field in favour of the one below?

internal/core/note_index.go
```diff
@@ -25,6 +25,12 @@ type NoteIndex interface {
 	// FindCollections retrieves all the collections of the given kind.
 	FindCollections(kind CollectionKind, sorters []CollectionSorter) ([]Collection, error)
 
+	// FindAllTags retrieves every occurrence of a tag from the index.
```
WhyNotHugo:
```suggestion
	// FindAllTags retrieves all tag associations with a given tag from the index.
```

internal/core/note_index.go
```diff
@@ -25,6 +25,12 @@ type NoteIndex interface {
...
+	FindTag(tagName string, sorters []TagSorter) ([]TagDetail, error)
+
+	// FindAllTags retrieves every occurrence of every tag from the index.
```
WhyNotHugo:
```suggestion
	// FindAllTags retrieves all tag associations of every tag from the index.
```

internal/core/note_parse.go
```diff
@@ -33,6 +33,8 @@ type NoteContent struct {
 	Body opt.String
 	// Tags is the list of tags found in the note content.
 	Tags []string
```
WhyNotHugo, line 35: Can we drop this field entirely in favour of the one below?

Or is that too big of a refactor, and best done separately?

internal/cli/cmd/tag.go
```diff
@@ -118,31 +156,50 @@ func (cmd *TagList) Run(container *cli.Container) error {
...
-
 	return err
 }
 
-func (cmd *TagList) tagTemplate() string {
-	format := cmd.Format
-	if format == "" {
-		format = "full"
+func (cmd *TagList) emitTagPositions(notebook *core.Notebook, container *cli.Container, count *int) error {
+
```
WhyNotHugo, line 122:
```suggestion
```

internal/adapter/sqlite/tags_dao.go
```diff
@@ -0,0 +1,180 @@
...
+				JOIN notes n
+				ON t.note_id = n.id
+				WHERE c.kind =  '` + string(core.CollectionKindTag) + `'
+				AND c.name = ?
+				GROUP BY c.name, n.title, n.path
+		`,
+	}
+}
+
+func parsePositions(positionsCat string) []int { // 1. Trim whitespace just in case the DB driver added padding
```
WhyNotHugo: Huh? This sounds like a bug in insertion.

internal/adapter/sqlite/tags_dao.go
```diff
@@ -0,0 +1,180 @@
...
+
+func (d *TagDAO) buildQuery(query string, sorters []core.TagSorter) string {
+	orderTerms := []string{}
+	for _, sorter := range sorters {
+		orderTerms = append(orderTerms, tagOrderTerm(sorter))
+	}
+
+	if len(orderTerms) > 0 {
+		query += "ORDER BY " + strings.Join(orderTerms, ", ") + "\n"
+
```
WhyNotHugo:
```suggestion
```

internal/cli/cmd/tag.go
WhyNotHugo, line 1: We should print JSON output for `--format=positions`, rather than JSONL. This avoids repeating all the data for each tag N times, and consumers can easily consume this output with any tool (including `jq`).

Edit: As an alternative: what do you think of using a separate command entirely (e.g.: `tagged-notes`) for this output?

internal/util/yaml/yaml.go
```diff
@@ -1,56 +1,226 @@
...
+	case yml.ScalarNode:
+		return walkScalar(node)
+	case yml.SequenceNode:
+		return walkSequence(node)
+	case yml.MappingNode:
+		return walkMapping(node)
+	case yml.AliasNode:
+		return walkNode(node.Alias)
+	}
+	return nil, nil // The cases above should be exhaustive.
```
WhyNotHugo:
```suggestion
	panic("Unreachable: cases above are exhaustive.")
```

WhyNotHugo (commented):
High level look good to me, only minor comments.

My main reservation at this point is the output format. I'm not a fan of this _only_ being available via JSONL. I'd rather this be JSON (which at least is easy to consume via scripts), but we can also consider having a separate command. Strictly speaking, this doesn't enumerate all tags, but all tag-associations, or tagged-notes.

almbrs:
I am having a problem now that after trying to make these changes:

```sh
~/work/zk dev > go test                                     
?       github.com/zk-org/zk    [no test files]
```
I have no idea why and I can't continue until tomorrow.

internal/cli/cmd/tag.go
ghost: > We should print JSON output for `--format=positions`, rather than JSONL. This avoids repeating all the data for each tag N times, and consumers can easily consume this output with any tool (including `jq`).
> 
> Edit: As an alternative: what do you think of using a separate command entirely (e.g.: `tagged-notes`) for this output?

Can you give me an example of the output that you would like for it to produce?

internal/core/note_parse.go
ghost: I would like to avoid making this change right now.

WhyNotHugo:
Nobody's ignoring you.

We're  human beings who have other things to deal with in life. Like: work, conferences, family, eating, sleeping, health, etc. Nobody exists solely for this project.

If I'd had time in the past days, I would have had a look at this 1100 LoC. But this isn't something I can quickly review in 15 minutes, and I haven't had the time.

internal/cli/cmd/tag_positions.go
WhyNotHugo, line 1: Nice, I like this layout for the command a lot better.

The output repeats each tag multiple times though, what do you think of instead of the following:

    {
      "tag": "report",
      "title": "2025-11-25 Report",
      "path": "journal/2025-11-25_1520.md",
      "position": [
        32
      ]
    },
    {
      "tag": "report",
      "title": "2025-12-01 Report",
      "path": "journal/2025-12-01_1525.md",
      "position": [
        35
      ]
    },

Producing:

    {
      "tag": "report",
      "notes": {
        "title": "2025-11-25 Report",
        "path": "journal/2025-11-25_1520.md",
        "position": [
          32
        ]
        "title": "2025-12-01 Report",
        "path": "journal/2025-12-01_1525.md",
        "position": [
          35
        ]
      }
    },

internal/cli/cmd/tag_positions.go
```diff
@@ -0,0 +1,51 @@
...
+func writeJSONL(w io.Writer, records []core.TaggedNoteDetail) error {
+    enc := json.NewEncoder(w)
+    enc.SetEscapeHTML(false) // optional
+    for _, r := range records {
+        if err := enc.Encode(r); err != nil { // Encode appends "\n" automatically
+            return err
+        }
+    }
+    return nil
+}
```
WhyNotHugo: Unused?

internal/adapter/markdown/markdown.go
```diff
@@ -163,51 +187,33 @@ func parseLead(body opt.String) opt.String {
...
-					res = append(res, s)
-				}
-			}
-			return res
-
-		} else {
-			return []string{}
-		}
+func resolveTags(tags []string, pos int) []core.Tag {
+	// here we assume that if len(tags) > 1 then these are colon separated tags.
```
WhyNotHugo: I'm not sure I follow this comment.

internal/core/tag.go
```diff
@@ -0,0 +1,67 @@
...
+func TaggedNoteSorterFromString(str string) (TaggedNoteSorter, error) {
+	orderSymbol, _ := utf8.DecodeLastRuneInString(str)
+	str = strings.TrimRight(str, "+-")
+
+	var sorter TaggedNoteSorter
+	switch str {
+	case "name", "n":
+		sorter = TaggedNoteSorter{Field: TagSortName, Ascending: true}
+	case "title", "t":
+		sorter = TaggedNoteSorter{Field: TagSortTitle, Ascending: false}
```
WhyNotHugo: One of these defaults to descending and the other to ascending. Is this intentional?

It seems a bit counter-intuitive.

internal/core/notebook.go
```diff
@@ -356,6 +363,18 @@ func (n *Notebook) NewCollectionFormatter(templateString string) (CollectionForm
...
+	if err != nil {
+		return nil, err
+	}
+	template, err := templates.LoadTemplate(templateString)
+	if err != nil {
+		return nil, err
+	}
+	return taggedNoteFormatter(template)
+}
+
```
WhyNotHugo: Unused?

internal/core/tag_format.go
```diff
@@ -0,0 +1,25 @@
...
+	"encoding/json"
+)
+
+type TaggedNoteFormatter func(tag TaggedNoteDetail) (string, error)
+
+func taggedNoteFormatter(template Template) (TaggedNoteFormatter, error) {
+	return func(tag TaggedNoteDetail) (string, error) {
+		return template.Render(tag)
+	}, nil
+}
```
WhyNotHugo: Unused too?

internal/adapter/markdown/markdown.go
```diff
@@ -163,51 +187,33 @@ func parseLead(body opt.String) opt.String {
...
-		for _, t := range findFMTags(key) {
-			// Trims any # prefix to support hashtags embedded in YAML
-			// frontmatter, as in Simple Markdown Zettelkasten:
-			// http://evantravers.com/articles/2020/11/23/zettelkasten-updates/
-			tags = append(tags, strings.TrimPrefix(t, "#"))
-		}
-	}
+func collectTags(tags []string, pos int) []core.Tag {
+	return resolveTags(tags, pos)
+}
```
WhyNotHugo: Could just call `resolveTags` directly.

internal/core/tag.go
ghost: This was copy-pasted from the `Collections` sorter. I didn't notice this until now.

WhyNotHugo:
Tesh tests need to be updated.

ghost:
I have rebased. Please let me know if anything else needs to change.

internal/util/yaml/yaml.go
```diff
@@ -1,56 +1,226 @@
...
+				i += size
 			}
+			return i
+		}
+		if src[i] == '\n' {
+			l++
 		}
-		v = m
+		i++
+	}
```
WhyNotHugo, line 82: Common convention is to have `i++` part of the `for`:

```suggestion
	for i := 0; i < len(src); i++ {
		if l == line {
			for c := 1; c < col; c++ {
				_, size := utf8.DecodeRune(src[i:])
				i += size
}
			return i
		}
		if src[i] == '\n' {
			l++
}
	}
```

internal/util/yaml/yaml.go
```diff
@@ -1,56 +1,226 @@
...
-	case map[string]any:
-		for k, v2 := range x {
-			x[k] = ConvertToJSONCompatible(v2)
+func collectTags(rootMapping *yml.Node, frontmatter *FrontmatterExtended) ([]core.Tag, error) {
+	rootMappingContent := rootMapping.Content
+	rootMappingContentLength := len(rootMappingContent)
+	parsedTags := []core.Tag{}
+	frontMatterBytes := frontmatter.Bytes
+	for i := 0; i < rootMappingContentLength; i += 2 {
+		res := []core.Tag{}
```
WhyNotHugo, line 145: This variable is never used (`res … :=` declared again below).

internal/util/yaml/yaml.go
```diff
@@ -1,56 +1,226 @@
...
+						if len(trimmedTag) > 0 {
+							parsedTags = append(parsedTags, core.Tag{Name: trimmedTag, Pos: tagOffset})
+						}
+						currentOffset = tagOffset + len(tag)
+					}
+				}
+			case yml.SequenceNode:
+				valNodeContent := valNode.Content
+				for _, seqItemNode := range valNodeContent {
+					if !(seqItemNode.Kind == yml.ScalarNode) {
```
WhyNotHugo, line 122:
```suggestion
					if seqItemNode.Kind != yml.ScalarNode {
```

internal/adapter/sqlite/tags_dao.go
```diff
@@ -0,0 +1,177 @@
...
+	"database/sql"
+	"fmt"
+	"strings"
+
+	"github.com/zk-org/zk/internal/core"
+	"github.com/zk-org/zk/internal/util"
+)
+
+// TaggedNoteDAO persists tag-note associations to the SQLite database.
+
```
WhyNotHugo, line 13:
```suggestion
```

WhyNotHugo (changes requested):
Just some minor feedback at this point, I think it's good to merge once these are addressed.

internal/util/yaml/yaml.go
WhyNotHugo: Ugh, I don't know why GitHub likes to mangle formatting so much.

PR #703, opened 2026-04-30
Fix wrong end position being found for links

### LINKED ISSUE #702: regression: Wrong link followed in 0.15.3
### Check if applicable

### Describe the bug

Since 0.15.3, the LSP's "jump to definition" often follows the wrong link.

The last three lines of my note are separate links. Following the first works properly. Following the other two takes me to the same destination as the first link, even though the point to different files.

### How to reproduce?

```
# Test

- [one](e3bo)
- [two](5qpt)
- [three](kwpn)
```

All three links lead to `e3bo.md`.

### zk configuration

https://git.sr.ht/~whynothugo/dotfiles/tree/c19aadfa558f5f0a2ce824fa84f7e82a557f96a3/item/home/.config/zk/config.toml

### Environment

```bash
zk 0.15.3
system: Linux 6.19.11-0-stable x86_64 Linux
```

WhyNotHugo:
When following an link via the LSP integration, the target for the first link is often used. When a link has no siblings, its end position is the last byte in the document.

The code uses NextSibling().Pos() for searching backwards to find the link's closing delimiter. This assumption holds when multiple inline elements are siblings within the same parent, but fails when each link is isolated in its own block with no next sibling.

Use the next node, regardless of whether it is a sibling.

Fixes: b514571d3c0d12d86e36bc87fd3ead2ee74d930e
Fixes: https://github.com/zk-org/zk/issues/702

### DIFF
--- a/internal/adapter/markdown/link_position.go
+++ b/internal/adapter/markdown/link_position.go
@@ -20,9 +20,9 @@ func GetLinkPosition(link *ast.Link, source []byte) *LinkPosition {
 		return nil
 	}
 
-	// Find the position of the next sibling node (or EOF)…
+	// Find the position of the next node (or EOF)…
 	searchStart := len(source)
-	if next := ast.Node(link).NextSibling(); next != nil {
+	if next := nextNodeInAST(link); next != nil {
 		searchStart = next.Pos()
 	}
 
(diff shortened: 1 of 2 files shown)


WhyNotHugo:
@tjex I'd appreciate if we can have a bugfix release with this one, since the jump-to-definition is completely broken in very common cases.

WhyNotHugo:
Thanks!

PR #704, opened 2026-05-01
Add MacPorts to README

mcookly:
MacPorts has had zk for a while now, so it would be good to give macOS users an alternative method to Homebrew. I've added it in alphabetical order (after Arch) since it's a smaller package manager than Homebrew and Nix.


WhyNotHugo:
Thanks

Issue #321, opened 2023-05-19
Support links in YAML frontmatter


bolt12:
I want to be able to relate notes without it appearing explicitly in the markdown text. To do this I think a design would be to put something like

```yaml
links:
  - [[{{filename}}]]
  - [[{{filename}}]]
```

in the YAML frontmatter

Would this be something that can be done?

mickael-menu:
Depends what you want to use it with. Following frontmatter links with the LSP should already work, I don't think using `--link-to` will though.

bolt12:
Exactly that's the what I'd want to do! Be able to relate multiple notes
without cluttering the content of the notes. I could always make it
explicit, but would be neat to be able to not to and just have that
information for querying purposes

mickael-menu:
I took a quick look, unfortunately it doesn't seem like the Markdown parser I'm using is able to extract the links in the YAML frontmatter. So that would require significant changes.

bolt12:
Thanks for considering my request!

alancunha26:
Maybe this could be achieved similar to what Obsidian is doing?

```yaml
links:
  - "[[file|alias]]"
  - "[alias](link)"
```

I'm migrating to ZK and have a lot of notes with links under the frontmatter, the navigation works as intended, but the `ZkBacklinks` of the nvim plugin don't list this links.

tjex:
Supported with #710

PR #710, opened 2026-05-05
Parse links to notes in frontmatter

### LINKED ISSUE #709: Parse links in frontmatter
### Description

This has been requested before (and I have shot it down before...): https://github.com/zk-org/zk/issues/321, https://github.com/zk-org/zk-nvim/issues/201

While links in frontmatter does not adhear to the standard markdown specification, I think this functionality has merit. Firstly, links to notes in frontmatter _can_ be regarded as metadata. The link describes a relationship that the note has.

Secondly, it is no doubt a big friction point for people moving from Obsidian where this is supported.

### Do I plan to submit a PR for this?

- [ ] No
- [x] Yes

tjex:
Resolves: #709 #321 

Parses links in frontmatter blocks allowing features such as `--link-to`, `--linked-by` and therefore `:ZkBacklinks` etc in `zk-nvim`.

Links do work in frontmatter arrays (also itemised arrays), but I can't get the test case to work in Go. I have a tesh case for it though. The Go test states it wants a `Rel:` type of `{ "down" }`...?

I'm not nearly as familiar with the code base as you @WhyNotHugo and am not sure if I've missed any relevant structural steps... Your review would be appreciated!

I'm also not sure if this is the best appraoch. I needed to pass the frontmatter values to `p.parseLinks()` again, as from what I understand the preceding call at line 78 `markdown.go` runs with frontmatter removed from the ast. There is likely a more elegant way of handling this than I know of(?)

### DIFF
--- a/internal/adapter/markdown/markdown.go
+++ b/internal/adapter/markdown/markdown.go
@@ -85,6 +85,12 @@ func (p *Parser) ParseNoteContent(content string) (*core.NoteContent, error) {
 		return nil, err
 	}
 
+	fmLinks, err := p.parseFrontmatterLinks(frontmatter)
+	if err != nil {
+		return nil, err
+	}
+	links = append(links, fmLinks...)
+
 	title, bodyStart, err := parseTitle(frontmatter, root, bytes)
 	if err != nil {
 		return nil, err
@@ -269,6 +275,22 @@ func (p *Parser) parseLinks(root ast.Node, source []byte) ([]core.Link, error) {
 		return ast.WalkContinue, nil
 	})
 	return links, err
+}
+
+func (p *Parser) parseFrontmatterLinks(frontmatter frontmatter) ([]core.Link, error) {
+	links := []core.Link{}
+	for _, val := range frontmatter.values {
+		valStr := fmt.Sprint(val)
+		fmRoot := p.md.Parser().Parse(
+			text.NewReader([]byte(valStr)),
+		)
+		fmLinks, err := p.parseLinks(fmRoot, []byte(valStr))
+		if err != nil {
+			return nil, err
+		}
+		links = append(links, fmLinks...)
+	}
+	return links, nil
 }
 
 func extractLines(n ast.Node, source []byte) (content string, start, end int) {
--- a/internal/adapter/markdown/markdown_test.go
+++ b/internal/adapter/markdown/markdown_test.go
@@ -660,15 +660,29 @@ func TestParseLinkInFootnote(t *testing.T) {
 }
 
 func TestParseMetadataFromFrontmatter(t *testing.T) {
-	test := func(source string, expectedMetadata map[string]any) {
-		content := parse(t, source)
-		assert.Equal(t, content.Metadata, expectedMetadata)
-	}
-
-	test("", map[string]any{})
-	test("# A title", map[string]any{})
-	test("---\n---\n# A title", map[string]any{})
-	test(`---
+	tests := []struct {
+		name     string
+		source   string
+		expected map[string]any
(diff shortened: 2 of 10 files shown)


internal/adapter/markdown/markdown.go
```diff
@@ -271,6 +277,22 @@ func (p *Parser) parseLinks(root ast.Node, source []byte) ([]core.Link, error) {
...
+		fmRoot := p.md.Parser().Parse(
+			text.NewReader([]byte(valStr)),
+		)
+		fmLinks, err := p.parseLinks(fmRoot, []byte(valStr))
+		if err != nil {
+			return nil, err
+		}
+		links = append(links, fmLinks...)
+	}
+	return links, nil
```
marcoklein, line 293: To avoid the collision with Neurons legacy behavior `[[[link]]]` and a frontmatter array of links this function could remove the `Rels`. Otherwise, it interprets it as a "Folgezettel" - however, as Folgezettel are rather part of the body I think it's fine to not allow Folgezettel in frontmatter.

It's not 100% clean I think because the Folgezettel gets parsed and then we remove it again but the easiest viable option I see right now.

```go
for i := range links {
    links[i].Rels = []core.LinkRelation{}
}
```

internal/adapter/markdown/markdown_test.go
```diff
@@ -687,6 +687,146 @@ Paragraph
...
+			Rels:         []core.LinkRelation{},
+			Snippet:      "[[wikilink]]",
+			SnippetStart: 0,
+			SnippetEnd:   12,
+		},
+	})
+
+	// FIX: Links in arrays get a test Rels: value of {"down"}...?
+	// Links do work in arrays, but not sure what to make of this in terms of
+	// testability.
```
marcoklein: Seems to be due to the legacy Neuron feature "Folgezettel" trigger an extra relation (e.g. the triple `[[[link]]]` ). Made it work with the code I commented on the `markdown.go` file.

internal/adapter/markdown/markdown.go
tjex: This does work, but I don't think a blanked overwrite of the links Rels type is the way to go. Also, a blank slice like that is the type for a regular https:// link, and not a wiki or markdown link.

The further problem here is what happens if there are both web urls and markdown/wiki links in frontmatter? 

Also, are we supporting Nueorn specific features in our code?.. I don't see that.

marcoklein: Yes, I agree... Just overwriting the types would not be super clean. Regarding the Neuron feature it's buried in the wikilink parser: https://github.com/zk-org/zk/blob/main/internal/adapter/markdown/extensions/wikilink.go#L116

tjex:
@marcoklein Got tests working :) Could you check if this feature works for you, both with markdown and wikilinks? Would be greatly appreciated.

internal/adapter/markdown/markdown_test.go
```diff
@@ -687,6 +687,145 @@ Paragraph
 	})
 }
 
+func TestParseLinksInFrontmatter(t *testing.T) {
+	test := func(source string, links []core.Link) {
+		content := parse(t, source)
+		assert.Equal(t, content.Links, links)
+	}
```
WhyNotHugo: Please use [TableDrivenTests](https://go.dev/wiki/TableDrivenTests) instead.

WhyNotHugo (approved):
Core logic LGTM, just one suggestions on tests.

marcoklein:
Thank you! Ran a few tests and worked locally. Looking forward to the release 🎸

PR #712, opened 2026-05-06
Migrate static site builder to Zensical

tjex:
This is oppinionated. But I feel that Zensical is worth a crack despite being very new.

The interface is really nice out of the box. And its a much simpler setup, maintenance and onboarding compared to Sphinx.

In-built versioning (loading the documentation of different release versions) is also on the roadmap (with an interim third party solution). That was one of the wishes of Mickael with the original docs site that we couldn't follow through on as versioned documentation was a bit janky with Sphinx. 

This PR should build and deploy out of the box, but let's see...


tjex:
Closing as the needed workflow needs to be in in dev _before_ creating a pull request...

PR #714, opened 2026-05-06
Test deployment of Zensical static docs site

tjex:
@WhyNotHugo None of this is touching the zk sourcecode and won't trigger a release as there's no tags. So it's easily revertable, but all the same, do you have any objections to the new static site direction? 

It's building, just need to merge to main to test deployment.

Looks like this:

image

image


WhyNotHugo:
Looks okay to me.

They're missing a proper search page, which sphinx does have. Search results are shown in a tiny pop-up that's like 15% of the screen, and it only shows the title of the pages which match, but not the actual matching text. E.g.:

image

Sphinx shows the full preview including the context of keywords:

image

WhyNotHugo:
Actually, the lack of a search results page is quite annoying.

Results don't show context, so you have to blindly click results and see if they're what you're looking for. But when you press back… it doesn't take you back to results, because there is no results page, you have to search again and pray you remember which position you were checking.

Might be worth asking upstream if they want to intend to implement proper searching, the current implementation is subpar.

tjex:
Hmm ok. Yeah, fair enough. I'll revert on dev and put this on ice. Lets see later how they go.

tjex:
Reverted at: https://github.com/zk-org/zk/commit/234bf25c2f79645f21fe48753a5b41c9d61dd12a

PR #724, opened 2026-05-16
LSP option to append links to selected text

tjex:
Currently when creating a note with text selection as context, the selected text
is replaced with link of the resultant note.

This PR adds an `append = true` option to our LSP, which will append the link to
the selected text instead of replacing it.

E.g.,

```
|Selected text for creation|.
```

```
Selected text for note creation [[link-to-note]]|.
```

Motivation here is that often the selected text is important context for the
current note. Completely replacing it with a link is only practical when the
title of the note is used as the link text. If a link format takes a different
form (like using the note filename as the link text) then replacing the selected
text with this link destroys the comprehension of the note.

The append option therefore provides these users with a smoother workflow when
performing text selection based note creation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Format: `<description> (by <contributor>, <pr number>)`
   @WhyNotHugo, 708)
 - Update strftime package, supporting `%g` and `%G` formats in the
   `{{format-date}}` helper (by @tjex, 723)
+- Option to append links to selected text, instead of replacing (by @tjex, 724)
 
 ### Fixed
 
--- a/docs/tips/editors-integration.md
+++ b/docs/tips/editors-integration.md
@@ -132,9 +132,11 @@ language-servers = ["zk"]
 
 <!-- prettier-ignore -->
 
-:::{note} This configuration disables Helix's
+:::{note}\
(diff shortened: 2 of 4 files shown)


WhyNotHugo:
When using markdown links, the text:

    Hello there

Is replaced by:

    [Hello there](new-note.md)

This is ideal IMHO, since it linkifies the selection and doesn't discard the text.

Do wiki-links not support a similar syntax? Something like:

    [[link-to-note Selected text for note creation]]

WhyNotHugo (approved):
Changes look good to me, although I would much prefer if the default behaviour for wiki-links and markdown-links were consistent. It seems like right now wiki-links have a subpar behaviour.

tjex:
The default behaviour is maintained, also for markdown links. So without passing `append = true`:

```
Hello there
```

Will still become

```
[Hello there](path.md)
```

And with wikiliks

```
[[Hello there]]
```

With `{append = true, title = "test title"}`

```
Hello there [test title](path.md)
```

Wikilinks

```
Hello there [[test title]]
```

WhyNotHugo:
Ah, I misunderstood the previous behaviour for wikilinks and thought it was swallowing the original text. My bad.

PR #735, opened 2026-06-07
Parallel update of links

Keluaa:
This PR brings the time to index the large test data (`make testdata`) from 7min to ~8sec.

The main performance bottleneck was the compilation of dynamic regexes. The patterns they matched were simple enough to be converted into normal string searches.

The second bottleneck was the systematic link update when adding a new note.
Now we put new notes in an array, and update all links together in parallel afterward.
When adding a single file (`zk new`) we do the same parallel update but for a single file.

### DIFF
--- a/internal/adapter/sqlite/note_index.go
+++ b/internal/adapter/sqlite/note_index.go
@@ -5,12 +5,12 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
-	"sync"
 
 	"github.com/zk-org/zk/internal/core"
 	"github.com/zk-org/zk/internal/util"
 	"github.com/zk-org/zk/internal/util/paths"
 	strutil "github.com/zk-org/zk/internal/util/strings"
+	"golang.org/x/sync/errgroup"
 )
 
 // NoteIndex persists note indexing results in the SQLite database.
@@ -188,8 +188,7 @@ func (ni *NoteIndex) batchFixExistingLinks(dao *dao, ids []core.NoteID, paths []
 	// which is a considerable amount of work. We do so without using all CPU cores at once.
 	maxWorkers := min(max(runtime.GOMAXPROCS(0)-2, 1), len(links))
 	linksPerWorker := len(links) / maxWorkers
-	firstError := make(chan error, 1)
-	var wg sync.WaitGroup
+	group := new(errgroup.Group)
 	for wid := range maxWorkers {
 		start := linksPerWorker * wid
 		end := linksPerWorker * (wid + 1)
@@ -197,29 +196,17 @@ func (ni *NoteIndex) batchFixExistingLinks(dao *dao, ids []core.NoteID, paths []
 			// The last worker does a bit more work
 			end += len(links) % maxWorkers
 		}
-		wg.Add(1)
-		go func(start, end int) {
+		group.Go(func() error {
 			for _, link := range links[start:end] {
 				err := fixLink(link)
 				if err != nil {
-					select {
-					case firstError <- err:
-					default:
-					}
+					return err
 				}
 			}
-			wg.Done()
-		}(start, end)
+			return nil
+		})
 	}
-
-	wg.Wait()
-	close(firstError)
-	select {
-	case err := <-firstError:
-		return err
-	default:
-		return nil
-	}
+	return group.Wait()
 }
 
 // linkMatchesPath returns whether the given link can be used to reach the
--- a/internal/adapter/sqlite/note_index.go
+++ b/internal/adapter/sqlite/note_index.go
@@ -1,11 +1,11 @@
 package sqlite
 
 import (
-	"path/filepath"
-	"regexp"
-	"strings"
-
 	"fmt"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"sync"
 
 	"github.com/zk-org/zk/internal/core"
 	"github.com/zk-org/zk/internal/util"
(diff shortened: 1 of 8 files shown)


internal/adapter/sqlite/note_index.go
```diff
@@ -134,62 +136,119 @@ func (ni *NoteIndex) Add(note core.Note) (id core.NoteID, err error) {
...
+			}
+			if matches {
+				bestTargetPath = path
+				bestMatch = i
+			}
 		}
 
-		matches, err := ni.linkMatchesPath(link, path)
-		if matches && err == nil {
-			err = dao.links.SetTargetID(link.ID, id)
```
WhyNotHugo, line 155: We were updating the link target for every match each time we found a better one? 🤦

internal/adapter/sqlite/note_index.go
```diff
@@ -134,62 +136,119 @@ func (ni *NoteIndex) Add(note core.Note) (id core.NoteID, err error) {
...
 		}
-		if err != nil {
-			return err
+
+		return nil
+	}
+
+	// Update the links in parallel: we must parse through the links of all notes,
+	// which is a considerable amount of work. We do so without using all CPU cores at once.
+	maxWorkers := min(max(runtime.GOMAXPROCS(0)-2, 1), len(links))
```
WhyNotHugo, line 189: Isn't this work I/O bound anyway?

No complains here, just thinking out loud if we really need to cap this.

internal/adapter/sqlite/note_index.go
```diff
@@ -134,62 +136,119 @@ func (ni *NoteIndex) Add(note core.Note) (id core.NoteID, err error) {
...
+	linksPerWorker := len(links) / maxWorkers
+	firstError := make(chan error, 1)
+	var wg sync.WaitGroup
+	for wid := range maxWorkers {
```
WhyNotHugo, line 192: I think you can use https://godocs.io/golang.org/x/sync/errgroup here. It simplifies some of the worker management, and handles cancellation of pending jobs in case of an error.

Thoughts?

internal/adapter/sqlite/note_index.go
```diff
@@ -134,62 +136,119 @@ func (ni *NoteIndex) Add(note core.Note) (id core.NoteID, err error) {
...
 func (ni *NoteIndex) linkMatchesPath(link core.ResolvedLink, path string) (bool, error) {
 	// Remove any anchor at the end of the HREF, since it's most likely
 	// matching a sub-section in the note.
-	href := strings.SplitN(link.Href, "#", 2)[0]
-
-	matchString := func(pattern string, s string) bool {
-		reg := regexp.MustCompile(pattern)
-		return reg.MatchString(s)
+	href := link.Href
+	if hashPos := strings.LastIndexByte(link.Href, '#'); hashPos != -1 {
```
WhyNotHugo: This previously split on the first `#`, but not splits on the last one, which doesn't sound correct.

internal/adapter/sqlite/note_index.go
```diff
@@ -134,62 +136,119 @@ func (ni *NoteIndex) Add(note core.Note) (id core.NoteID, err error) {
...
 	matches := func(href string, allowPartialHref bool) bool {
 		if href == "" {
 			return false
 		}
-		href = regexp.QuoteMeta(href)
-
-		if allowPartialHref {
-			if matchString("^(.*/)?[^/]*"+href+"[^/]*$", path) {
-				return true
-			}
```
WhyNotHugo, line 186: ~~This matches if the filename (not the whole path) starts with `href`, the new implementation does not prioritise these matches.~~

Edit: right, I've always thought "targets matching this rule take precedence over those matching following rules, but we're not applying each rule in parallel: all these rules apply in sequence to the same note.

Your changes are correct, I'm just baffled that I always reasoned about this all wrong.

This regex match has always been completely redundant.

internal/adapter/sqlite/note_index.go
```diff
@@ -134,62 +136,119 @@ func (ni *NoteIndex) Add(note core.Note) (id core.NoteID, err error) {
...
+		pos := strings.Index(path, href)
+		if allowPartialHref && pos != -1 {
+			// Match if 'href' is anywhere in 'path'
+			return true
 		}
+		// Match only if 'href' prefixes 'path', and 'path' isn't a directory named "href/"
+		// e.g. with 'href="dir"', match "dir/file", "dir", or "dir2", but not "dir/"
+		return pos == 0 && !(len(path) == len(href)+1 && path[len(path)-1] == '/')
```
WhyNotHugo: The previous rule matched two cases:

- `href` is at the beginning of path and had no slash anywhere after it.
- `href` is at the beginning and has a slash immediately after it.

I don't think this replacement does the same.

WhyNotHugo (commented):
Thanks for tackling this.

It's hard to overstate how pleased I am with this set of changes. The improvement is massive.

Local test results:

## Before:

Full index:

```
Indexed 3402 notes in 3m48s
  + 3402 added
  ~ 0 modified
  - 0 removed
zk index  298.84s user 18.38s system 139% cpu 3:47.90 total
```

Incremental index:

```
Indexed 3402 notes in 0s
  + 0 added
  ~ 2 modified
  - 0 removed
zk index  0.03s user 0.01s system 73% cpu 0.052 total
```

## After

Full index:

```
Indexed 3402 notes in 3.5s
  + 3402 added
  ~ 0 modified
  - 0 removed
./zk index  11.00s user 0.43s system 312% cpu 3.658 total
```


Incremental index:

```
Indexed 3402 notes in 0s
  + 0 added
  ~ 0 modified
  - 0 removed
./zk index  0.07s user 0.02s system 126% cpu 0.070 total
```

The difference is huge, 3.5m into 3.6s.

I only have one minor observation where I think we handle two edge cases different, but overall looks great.

Keluaa:
The link matching has been fixed to match the original behaviour (that I misinterpreted...).

Following your suggestion I switched to using errgroup for parallel work, as it does clean up the code quite a bit.

WhyNotHugo (approved):

WhyNotHugo:
Thanks!

tjex:
Really awesome, thank you very much! 👏

PR #736, opened 2026-06-16
Filter LSP note completions from config

wrvsrx:
## Description

Adds a new `[lsp.completion] note-filter` setting to limit which notes are offered for LSP link completion.

Related discussion: https://github.com/zk-org/zk/discussions/329

The filter uses the same CLI filtering syntax as `zk list`, including support for named filters from `[filter]`.

Example:

```toml
[lsp.completion]
note-filter = "--tag project --sort modified-"
```

or:

```toml
[filter]
completion = "--tag project --exclude archive"

[lsp.completion]
note-filter = "completion"
```

## Changes

- Adds `note-filter` parsing to LSP completion config.
- Reuses CLI filtering parsing via a new `cli.ParseFiltering` helper.
- Applies the parsed filter before building LSP link completion items.
- Documents the new setting in LSP config docs and default config comments.
- Adds coverage for config parsing and filtered LSP completion.

## Testing

```sh
make test
make tesh
```

`make tesh` passes. `make test` currently fails in the existing `internal/adapter/term` ANSI color styling tests in this environment; the packages touched by this change pass with:

```sh
go test -tags fts5 ./internal/core ./internal/cli ./internal/adapter/lsp
```


internal/adapter/lsp/server_test.go
```diff
@@ -168,6 +169,16 @@ func TestServer_buildInvokedCompletionList(t *testing.T) {
...
+
+	t.Run("Filter link completion notes from config", func(t *testing.T) {
+		notebook.Config.LSP.Completion.NoteFilter = opt.NewString("--tag project")
+
+		items, err := server.buildInvokedCompletionList(notebook, doc, protocol.Position{Line: 3, Character: 2})
+
+		assert.Nil(t, err)
+		assert.Equal(t, len(items), 1)
+		assert.Equal(t, items[0].Label, "Item 3 that contains spaces")
+	})
```
tjex: How come the test for this new feature is included in its own `t.Run()` and not structured as its own table?

internal/cli/filtering.go
```diff
@@ -47,22 +47,31 @@ type Filtering struct {
 	ExactMatch bool `kong:"hidden,short='e'" json:"exactMatch"`
 }
 
+// ParseFiltering parses a shell-like filtering string into Filtering.
+func ParseFiltering(filter string) (Filtering, error) {
```
tjex: ParseFiltering isn't as grammatically sound as ParseFilter. I would say to change to ParseFilter

tjex (changes requested):
Thanks for this ❤️ A few comments to review.

internal/adapter/lsp/server_test.go
wrvsrx: Good point. I've moved this case into the existing test table and added fields for the note filter and expected labels. Each case now sets the filter explicitly, which also prevents configuration state from leaking between cases.

internal/cli/filtering.go
wrvsrx: Agreed. I've renamed ParseFiltering to ParseFilter and updated its call sites.

tjex (approved):
Thanks for the continual work on this. It's a really useful addition!

PR #741, opened 2026-06-23
Prune excluded directories from the index walk

### LINKED ISSUE #728: exclude patterns should skip directories and not only files
### Description

Hello,

Thanks for this tool. I noticed that the exclude patterns in the `[note]` section of the config are applied per-file, and there is no way to skip entire directories. So, even if I want to exclude completely `excluded-dir` (with all the subdirectories), `zk` will still traverse and check file by file. 

Motivation: I'm experiencing a performance issue with indexing. I think skipping directories would be a great speedup (e.g., node_modules). I have my vault in a network-mounted filesystem, so traversing is expensive.

### Do I plan to submit a PR for this?

- [x] No
- [ ] Yes

ehsash:
## Summary

Fixes #728.

Exclude globs are currently only matched against files, so an excluded
directory is still walked and filtered file by file. On notebooks that contain
large ignored trees (asset directories, vendored repos, ...) the indexer spends
most of its time `lstat`-ing files it then discards, which makes indexing — and
every `zk` invocation that reindexes — slow.

This makes the walker **prune** a directory matched by an exclude glob, with
`filepath.SkipDir`, instead of descending into it.

## Changes

- `shouldIgnorePath` now takes an `isDir bool`:
  - a directory skips the note-extension check (it has no `.md` extension) and
    is matched against the exclude globs alone;
  - both the bare `foo` and the subtree `foo/**` forms prune the directory.
- `paths.Walk` prunes a directory rejected by `shouldIgnorePath` with
  `filepath.SkipDir` (the notebook root is never pruned).
- Backward compatible: file-only globs (`*.png`, `dir/*`) match exactly as
  before; only directory matches now prune.

## Tests

- `internal/util/paths/walk_test.go`: a new test asserts the walker never even
  queries paths *inside* a pruned directory — proving the subtree is skipped,
  not filtered file by file.
- `tests/issue-728.tesh`: end-to-end test covering both the `foo` and `foo/**`
  forms.
- `gofmt` clean; `go test ./...` passes.

## Impact

On a ~12k-note notebook sitting next to a large (40+ GB, ~170k-entry) asset
directory, excluding that directory takes indexing from re-walking every asset
entry on each index to skipping the tree entirely.

### DIFF
--- a/internal/core/note_index.go
+++ b/internal/core/note_index.go
@@ -135,7 +135,13 @@ func (t *indexTask) execute(callback func(change paths.DiffChange)) (NoteIndexin
 		}
 
 		for _, ignoreGlob := range group.ExcludeGlobs() {
-			matches, err := doublestar.PathMatch(ignoreGlob, path)
+			var matches bool
+			var err error
+			if isDir {
+				matches, err = globPrunesDir(ignoreGlob, path)
+			} else {
+				matches, err = doublestar.PathMatch(ignoreGlob, path)
+			}
 			if err != nil {
 				return true, fmt.Errorf("failed to match exclude glob %s to %s: %w", ignoreGlob, path, err)
 			}
@@ -219,4 +225,27 @@ func (t *indexTask) execute(callback func(change paths.DiffChange)) (NoteIndexin
 		return stats, fmt.Errorf("indexing failed: %w", err)
 	}
 	return stats, nil
+}
+
+// globPrunesDir reports whether the exclude glob lets the whole directory be
+// pruned from the index walk, as opposed to only matching some of its entries.
+//
+// A directory is pruned only when the glob excludes its entire subtree: either
+// by naming the directory itself (e.g. "dir") or by covering its descendants
+// (e.g. "dir/**"). A glob that only matches the directory's immediate children
+// (e.g. "dir/*") must not prune it, or notes nested deeper — which the glob
+// never matches — would be dropped along with it.
+func globPrunesDir(glob, dir string) (bool, error) {
+	matches, err := doublestar.PathMatch(glob, dir)
+	if err != nil || !matches {
+		return false, err
+	}
+	// The glob names the directory exactly.
+	if glob == dir {
+		return true, nil
+	}
+	// The glob also matches a path below the directory, so it covers the whole
+	// subtree (as "dir/**" does) rather than only the immediate children (as
+	// "dir/*" does, which never matches a grandchild).
+	return doublestar.PathMatch(glob, dir+"/x/y")
 }
--- a/internal/core/note_index.go
+++ b/internal/core/note_index.go
@@ -3,6 +3,7 @@ package core
 import (
 	"fmt"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/bmatcuk/doublestar/v4"
@@ -113,7 +114,7 @@ func (t *indexTask) execute(callback func(change paths.DiffChange)) (NoteIndexin
 	}
 	ignoredFiles := []IgnoredFile{}
 
-	shouldIgnorePath := func(path string) (bool, error) {
+	shouldIgnorePath := func(path string, isDir bool) (bool, error) {
 		notifyIgnored := func(reason string) {
 			ignoredFiles = append(ignoredFiles, IgnoredFile{
 				Path:   path,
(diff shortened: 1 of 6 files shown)


internal/core/note_index.go
```diff
@@ -126,13 +127,22 @@ func (t *indexTask) execute(callback func(change paths.DiffChange)) (NoteIndexin
...
 		}
 
 		for _, ignoreGlob := range group.ExcludeGlobs() {
-			matches, err := doublestar.PathMatch(ignoreGlob, path)
+			glob := ignoreGlob
+			if isDir {
+				// A directory is excluded when a glob targets the directory
+				// itself ("foo") or its whole subtree ("foo/**").
+				glob = strings.TrimSuffix(glob, "/**")
+			}
```
WhyNotHugo: I'd prefer to skip this special `**` extension and just use regular glob matching (mainly for some added simplicity). This extension would only be useful if we supported things like `assets/**/*.css`, but we won't implement such things.

internal/util/paths/walk.go
```diff
@@ -24,18 +26,31 @@ func Walk(basePath string, logger util.Logger, notebookRoot string, shouldIgnore
...
+				logger.Println(err)
+				return nil
+			}
+
 			if info.IsDir() {
 				if isHidden && !isNotebookRoot {
 					return filepath.SkipDir
 				}
+				// Prune excluded directories so the walk never descends into
+				// large ignored trees (e.g. asset or vendored directories).
```
WhyNotHugo: Too verbose and specific:


```suggestion
				// Prune excluded directories.
```

WhyNotHugo:
Other than those two details, looks good.

internal/core/note_index.go
ehsash: Good remark. Noted and fixed.

internal/util/paths/walk.go
ehsash: Fixed.

tjex:
@ehsash Added one change to docs. The inner workings of how a walk is happening doesn't need to be explained in user facing documentation. 

Also, it was written that a directory can be excluded with `assets` (now changed to `your-directory` for more clarity). 

Shouldn't it be `your-directory/` to ignore the directory, and not `your-directory`? Is this how this fix is working, and the documentation missed this detail? Otherwise, what if there is a file with the same name as an excluded directory?

WhyNotHugo:
> Otherwise, what if there is a file with the same name as an excluded directory?

If you have a file named `assets` and a directory named `assets`, the file doesn't have the proper extension, and would get ignored anyway.

If you have a directory called `assets.md` a file with the same name, configured extension `.md` and are ignoring `assets.md`, then yes, both get ignored.

tjex:
Ah yep, of course. That makes sense.

@ehsash Not sure if you've seen, but the tesh case is failing. Should I look at it, or you good to look at it?

ehsash:
> Ah yep, of course. That makes sense.
> 
> @ehsash Not sure if you've seen, but the tesh case is failing. Should I look at it, or you good to look at it?

I'll have a look tomorrow afternoon and take the other remarks in account. You'll have a new commit in the PR that fixes this :-)

ehsash:
Your previous requests should be taken in account, this time. Tell me @WhyNotHugo and @tjex if there are anything that needs to be fixed, still.

ehsash:
@WhyNotHugo @tjex : pushed the fix for the failing `config-note-exclude.tesh`, rebased on `dev` (the CHANGELOG conflict is resolved). Both of Hugo's inline remarks are in as well.

On the test failure, since it touches the directory-matching logic Hugo commented on, it's worth explaining what I landed:

The regression was that plain glob matching over-prunes directories. `config-note-exclude` excludes with `['*rang*', 'dir/*']` over this tree:

    banana.md
    orange.md
    dir/orange.md
    dir/subdir/apple.md

`dir/*` is meant to exclude the *immediate* entries of `dir`, so `dir/subdir/apple.md` should stay indexed; it doesn't match `dir/*`. But when we match a directory against the globs, `PathMatch("dir/*", "dir/subdir")` is true, so the walker prunes `dir/subdir` and drops `apple.md` with it, even though the glob never matched that file.

So a directory can't be pruned just because its path matches a glob; pruning removes the whole subtree, which is only correct when the glob excludes that whole subtree. That's what `globPrunesDir` checks: prune only when the glob names the directory itself (`dir`) or covers its descendants (`dir/**`), not when it merely matches an immediate child (`dir/*`).

I kept it to regular glob matching per your earlier point, Hugo — no `/**` string surgery. The extra check is two `PathMatch` calls: one to confirm the glob matches the directory, one to confirm it also reaches below it (so `dir/*` is excluded but `dir/**` is not). Files are unchanged; they still go through a single `PathMatch`.

Happy to shape it differently if you'd prefer another formulation.

tjex:
Great, thanks for coming back to this. Doing some double checks on my end now.

tjex:
This results in nothing being ignored. 

```
exclude = [
   "foo/",
]
```
I think we should support that syntax as its supported in gitignore, and therefore enough people will write this out of habbit.

Also, this also results in nothing being ignored:

```
exclude = [
   "foo/*/bar",
]
```

Which is also a supported gitignore syntax. But should we support that? Or just require that people be more specific:

```
exclude = [
   "foo",
   "foo/bar",
]
```

WhyNotHugo:
I'm not sure I agree: git rules are **really** complex once you start getting into the finer details and edge cases (don't forget `!rules` and ordering). I'm not convinced that we need them. Even if we did, I don't think we block hold this particular PR on such a feature: the focus here is much narrower than that.

ehsash:
Extending the syntax is feasible. That's a design question only you, the people already in the project, can decide, as it's a question of general design and convention chosen for the codebase. You tell me :-). I'll implement the way you prefer to get it. I kind of agree with WhyNotHugo: if we go for gitignore convention, it must be fully implemented properly so that we don't get a half-implemented syntax that doesn't account for all the edge-case the same way git does.

tjex:
Fair fair. Alright, then I'll merge this. An easy sign if `dir/` is needed at least, is if we start to get 'bug reports' that directories aren't being ignorred. It will be a simple fix then anyway all the same.

Thanks for your contribution @ehsash. Really appreciate the detail you put into this. Hope to see more from you :)

PR #747, opened 2026-07-10
Drop dependency testfixtures

WhyNotHugo:
Dependency `testfixtures` pulls many transitive dependencies, and is only used in internal/adapter/sqlite/transaction_test.go.

These tests don't need any of the cross-database abstraction (zk only supports sqlite). They also run on a clean in-memory database, so also don't need table cleanups, sequence resets, or all the added drivers.

Replace the dependency with a single file which implements the necessary logic. Drop the the dependency and all its dependency tree, including multiple database drivers, networking and crypto packages and systemd bindings.


WhyNotHugo:
The fact that this adds less lines of code than the amount of lines removed from the lockfile says a lot.

jurica (approved):
LGTM

PR #750, opened 2026-07-16
fix: Preserve all NoteIndex fields in Commit

ehsash:
`NoteIndex.Commit` runs its callback against a fresh `NoteIndex` built from the
receiver — but from only three of its five fields:

    func (ni *NoteIndex) Commit(transaction func(idx core.NoteIndex) error) error {
        return ni.commit(func(dao *dao) error {
            return transaction(&NoteIndex{
                db:     ni.db,
                dao:    dao,
                logger: ni.logger,
            })
        })
    }

`notebookPath` and `extension` are left at their zero values, so the index
passed to the transaction is not equivalent to the one it wraps. This matters
because the whole indexing pass runs inside `Commit`
(`Notebook.IndexWithCallback` → `indexTask.execute`), so the wrapped index is
what does all the work.

I couldn't find a case where this changes behaviour on `dev` today, and I don't
want to overstate it:

- `notebookPath` is only read by `relNotebookPath`, where it appears in both
  the `Join` and the `Rel` and cancels out — the result is the same for `""`
  and the real path.
- `extension` is read from the `NoteDAO`, which is built by the outer index and
  reused, not taken from the wrapped index.

The bug is therefore latent rather than observable in the current codebase. It
still seems worth fixing: the zero value of each field is itself valid, so code
later added to the transaction path that reads `notebookPath` or `extension`
from the wrapped index would misbehave silently, with nothing to point at.

The fix carries every field so the wrapped index behaves like its receiver, plus
a test asserting the fields survive `Commit`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ Format: `<description> (by <contributor>, <pr number>)`
 - Indexing made significantly more performant (by @Keluaa, 735)
 - Support filtering by date and time with `"<date> <time>"` instead of
   `<date>T<time>` only (by @tjex, 743)
+- `NoteIndex.Commit` now passes a complete `NoteIndex` to its transaction
+  instead of one missing the notebook path and note extension (by @ehsash)
 
 ## 0.15.5
 
--- a/internal/adapter/sqlite/note_index.go
+++ b/internal/adapter/sqlite/note_index.go
@@ -345,10 +345,15 @@ func (ni *NoteIndex) Remove(path string) error {
 // Commit implements core.NoteIndex.
 func (ni *NoteIndex) Commit(transaction func(idx core.NoteIndex) error) error {
 	return ni.commit(func(dao *dao) error {
+		// Carry every field over so the index handed to the transaction
+		// behaves like its receiver; a partial copy leaves zero-value fields
+		// that silently diverge from the outer index.
 		return transaction(&NoteIndex{
-			db:     ni.db,
-			dao:    dao,
-			logger: ni.logger,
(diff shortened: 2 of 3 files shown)


WhyNotHugo:
If the fields are always mis-populated and never really used, I wonder if we can drop them. Clearly them having correct values is not necessary for the whole program to function correctly.

ehsash:
> If the fields are always mis-populated and never really used, I wonder if we can drop them. Clearly them having correct values is not necessary for the whole program to function correctly.

The fields cannot be dropped: they are actively used by the outer `NoteIndex`:

- `notebookPath` at lines 271 and 285 (`linkMatchesPath` and `relNotebookPath`)
- `extension` and `logseqCompat` at line 423, fed into `NewNoteDAO` inside the low-level `commit`

"Latent" referred specifically to the wrapped index built inside `Commit`. For that inner instance, the wrong values happen not to bite today:

- `notebookPath = ""` cancels in the path arithmetic because `SourcePath` is already notebook-relative, so both operands in `filepath.Rel` end up relative and the join-then-rel round-trips cleanly.
- `extension` / `logseqCompat` flow through the dao built by the outer commit, not re-read from the inner struct's fields (because the inner `NoteIndex` has `ni.dao != nil`, so the `if ni.dao != nil` branch in commit fires and skips `NewNoteDAO`).

So the accidental correctness is structural, not random — but it is accidental. The fix makes the wrapped index a faithful replica of its receiver, which is what the interface contract requires. Dropping the fields would be a different (larger) refactor: turn them into parameters threaded through the call sites that need them. That is a valid design, but it is not what this PR is doing.

PR #751, opened 2026-07-16
Update go-crypto

WhyNotHugo:
Fixes a bunch of security issues which most likely don't affect us. Go 1.24 is no longer maintained upstream.

See: https://github.com/zk-org/zk/pull/746

...

Burn after reading:

- Link related PRs
- Run tests (`make test` and `make tesh`)
- Update documentation where appropriate
- Add your contribution along with this PR number to `CHANGELOG.md`

PR #754, opened 2026-07-22
update to go 1.26

jurica:
Bump outdated actions and go to 1.26 as required for https://github.com/zk-org/zk/pull/751#issuecomment-5043488268

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -172,9 +172,9 @@ jobs:
         id: vars
         run: echo "version=$(git describe --tags --match v[0-9]* 2> /dev/null)" >> $GITHUB_OUTPUT
       - name: Setup go
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@v7
         with:
-          go-version: 1.23
+          go-version: 1.26
       - name: Build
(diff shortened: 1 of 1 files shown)


WhyNotHugo:
~~https://github.com/zk-org/zk/pull/751 only requires 1.25, not 1.26.~~

Never mind, this only affects runners and won't affect downstream

.github/workflows/build.yml
```diff
@@ -165,16 +165,16 @@ jobs:
...
           fetch-depth: 0
       - name: Set vars
         id: vars
         run: echo "version=$(git describe --tags --match v[0-9]* 2> /dev/null)" >> $GITHUB_OUTPUT
       - name: Setup go
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@v7
         with:
-          go-version: 1.23
+          go-version: 1.26
```
julio-lopez: nit: this can be derived from the `go.mod` file with `go-version-file: 'go.mod'`. Would that not work here? Same below.


```suggestion
          go-version-file: 'go.mod'
```

jurica: You‘re right, good catch! I wasn‘t aware of that option.
Feel free to make that change, I won‘t have my pc available till sunday/monday. Otherwise I‘ll make the adjustments after I return.

PR #755, opened 2026-07-25
Release v0.15.6

tjex:


WhyNotHugo:
A bit tangential, but: why can't we just tag off `dev`? It's not like we're going to have LTS relases off the main branch or something.

WhyNotHugo (approved):
Please merge ff, don't squash, so histories align.

tjex:
I did have a description explaining the rationale here, but I hit some keybind (qutebrowser) and lost it. I was too tiered and annoyed to write again. And I did want to get your thoughts on it. So here I am in the morning again..

The release process would then be:

- Merge dev into main
- Pull main
- Tag
- Push tag

The points to do this merge from dev to main for releases were:

- Provides more accessible and granular documentation of changes of releases.
- Allows others to discuss whether a release should be delayed because a certain bug fix / feature / PR should be merged.
- Pulling from main after successful merge means we will tag a commit that aligns with the docs site being published.

I agree that it seems a bit cumbersome / redundant in some ways.

You have way more experience with releasing in open source. Happy to change / update to what you see fit.

The simplest alternative would be to remove the push block rule from main, and just merge from dev locally, tag and push. This also avoids the annoying af problem of the rebase/merge button keeping the state of the last used option and creating frustrating git history cleanup work... But I'm also not sure if that's the better solution.

Could it be that we don't even need a dev branch?..

tjex:
Oh right, the strat just defaults back to squash on every PR...

https://github.com/orgs/community/discussions/20583

tjex:
Ok, no. This isn't good. There is no merge --ff option via github.


Let me know your thoughts here as to what the better approach is so that releases aren't as error prone. 

I would say two options:

1. Remove push restrictions on main, `merge --ff` dev into main, tag, push.
2. Do away with the dev branch entirely, and merge PRs directly into main. Then just tag and push tags.

tjex:
Changed the default branch to `main`. Makes most sense given main has no function apart from showing the history up until the latest release. The docs site is built on merges to main, but that can also just be triggered by pushing tags, so it's properly in tandem with a release.

WhyNotHugo:
I branch X has PR 123, and PR 123 has all checks passing, then you can run `git push origin X:main`, and GitHub allows it (since all checks are green), and also marks the PR as merged.

That's what I typically do: it bothers me to no end that GitHub doesn't have a `--ff-only` mode, which is the most sensible way to integrate a large branch into another.

This also works really well if PR 124 is a continuation of 123 and includes all the commits in 123, since it doesn't need a rebase.

That said, for our use case I think we should just work on a single branch (`main`/`dev`) and tag releases off it. Having separate branches is useful when you have a "live" system deployed from main and might need hotfixes. Or similar workflows which typically are a result of higher volumes of manpower and work.

tjex:
Great, thanks for the explanation. That makes sense needing a main on the grounds of having a live system deployed from it :+1:

Issue #740, opened 2026-06-20
Windows Defender flags release binary as Trojan (recurrence of #717)


sh1Nome:
### Check if applicable

- [x] I have searched the existing issues (**required**)
- [x] I'm willing to help fix the problem and contribute a pull request

### Describe the bug

This might be a false positive, but I'm filing this issue to make the team
aware. Sorry for the noise if so!

Microsoft Defender detects the v0.15.5 Windows release binary as a Trojan when
downloading it from the GitHub releases page. This is a recurrence of #717.

Additionally, building from source locally on Windows does not trigger Defender.
The v0.15.4 build used runner image [`win25/20260413.84`](https://github.com/zk-org/zk/actions/runs/25319059786/job/74223245298) while v0.15.5 used
[`win25/20260525.149`](https://github.com/zk-org/zk/actions/runs/27133506187/job/80080071054).

A screenshot of the Windows Defender log will be attached.

<img width="714" height="218" alt="Image" src="https://github.com/user-attachments/assets/7f1ba0aa-f522-4783-b689-2bf919c5049e" />

### How to reproduce?

1. Download `zk-v0.15.5-windows-x86_64.tar.gz` from the GitHub releases page.
2. Windows Defender flags and removes the file.

### zk configuration

```toml
N/A — the issue occurs before zk can be set up.
```

### Environment

```bash
zk 0.15.5
Windows 11 Home
```

WhyNotHugo:
My main concern is whether this is a false positive or not. I don't really know how to ascertain this.

A comment in https://github.com/zk-org/zk/issues/717 mentions that signing the binary prevents it being flagged by some heuristics. Looks like this can be done for free via https://signpath.org/

sh1Nome:
I ran the binary through VirusTotal and found that 14 out of 59 engines flagged it.  
I'm not sure how useful this information is for diagnosing or resolving the issue,  
but I hope it provides some additional context.  
https://www.virustotal.com/gui/file/64858cac1e8a6a3a486ef0cc6df1a4c6148e9766f1361e5d750eade3cb045b57/

jurica:
This was on and off on my mind the last weeks and I cannot figure out a sane way to handle it properly, particularly with the myriad of different environments people run.
Few days ago I stumbled over [gitbox](https://github.com/LuisPalacios/gitbox) and it's warning disclaimer, stating that it's an open source project and that the binaries available are not signed and might cause issues with built in OS protection.
[...]

Issue #717, opened 2026-05-09
File downloaded from github repo releases page but infected with Trojan:Win32/Kepavll!rfn


duytrung:
### Check if applicable

- [x] I have searched the existing issues (**required**)
- [x] I'm willing to help fix the problem and contribute a pull request

### Describe the bug

Hello developers,

I've updated my zk to 0.15.4. However I tried several times downloaded from github repo and extracted to the folder but the windows security reported the file was infected:

<img width="633" height="580" alt="Image" src="https://github.com/user-attachments/assets/1e7d367c-22a7-4005-b876-ee82cc822eeb" />

Could you please inspect the issue?

Thank you very much

### How to reproduce?

1. Download latest release 0.15.4 from releases page.
2. Extract to folder.
3. Windows Security removed the file and displays a message that the file was infected.

### zk configuration

```toml
(No config)
```

### Environment

```bash
zk 0.15.4.
Windows 11 Professional
```

tjex:
...that's weird.

I mean, I can't see how something like this could come from our side. The action to build the windows binary is running on the GitHub provided Windows runner. We're simply checking out the repository and building with go.

We're not using our own runners or images or doing any kind of dynamic additions during compilation. 

Could the Windows runner be compromised...? That would be embarasing with Microsoft owning GitHub and all...

https://github.com/zk-org/zk/blob/0b8e6aef53332fef3ba68a92298aa4299cc54f90/.github/workflows/build.yml#L218-L242

tjex:
The [windows runners will be updated from June 8th](
https://github.blog/changelog/2026-02-05-github-actions-early-february-2026-updates/#windows-server-2025-with-visual-studio-2026-image-now-available-for-github-hosted-runners), so lets see if that solves the issue.

jurica:
AFAIK this check on Windows uses heuristics and scans the binary data on disk for similarities with known malware. If it detects something it thinks to be suspicious the file is being marked accordingly. To my knowledge it's required to sign the binary in order to prevent that reliably, which unfortunately is out of reach here.
Most likely this is just coincidence and bad luck and should not happen again with the next build or release (as the binary data produced will be different from what it is now).

@duytrung if you have the possibility you can compile zk yourself on windows.

tjex:
@jurica would it be best here just to create another release? i.e, v0.15.4a or otherwise?

jurica:
[...]

PR #756, opened 2026-07-26
Add a disclaimer about the pre-built binaries to the README

sh1Nome:
Addresses #740, where the Windows binary from the release page is flagged and
quarantined by Windows Defender.

Following the direction suggested in that discussion, this adds a short
disclaimer to the README and points Windows users to the existing
build-from-source instructions.

Happy to adjust the wording if you would prefer it framed differently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,10 +54,10 @@ be released.
 pre-built binaries for macOS, Linux and Windows.
 
 > [!WARNING]
-> The pre-built binaries are not code-signed. Windows Defender and other
-> security tools have flagged the Windows binary as malicious. Building `zk`
-> yourself is currently the recommended workaround on Windows — see
-> [Build from scratch](#build-from-scratch). See
+> The pre-built binaries are not code-signed. In the past Windows Defender and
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -51,7 +51,14 @@ be released.
 ## Install
 
 [Check out the latest release](https://github.com/zk-org/zk/releases) for
-pre-built binaries for macOS and Linux (`zk` was not tested on Windows).
+pre-built binaries for macOS, Linux and Windows.
+
+> [!WARNING]
+> The pre-built binaries are not code-signed. Windows Defender and other
```
jurica: I would suggest to change the second sentence to: `In the past Windows Defender and other security tools have flagged some versions of the Windows binary as malicious.`
I thought not every release is being flagged, or am I missing something?

sh1Nome: Applied, thanks for catching this. You're right that not every release is affected, the only reports I know of are #717 and #740.

WhyNotHugo:
LGTM

jurica (approved):

PR #768, opened 2026-09-23
Honor the working dir group in edit --interactive

### LINKED ISSUE #767: edit --interactive's new note binding ignores the config group of the working directory
### Check if applicable

- [x] I have searched the existing issues

### Describe the bug

When running `zk edit --interactive` and triggering the "create a note" fzf binding (Ctrl-E by default), the new note is always created with the notebook root's config, regardless of which directory you're in.

I structure my notebook using only groups, each with their own template. I don't place anything at the root level, but I do have a `default.md` template assigned to root notes without actually defining this template. This has caught times when something didn't correctly get added to a group when I go to create it. Exactly like this!

Creating a note from the interactive editor fails with:

```
zk: error: new note: cannot locate template at default.md
Found 0 notes.
```

**What I expected:** since I'm inside `journal/`, a directory matched by the `journal` group, the created note should use that group's config — the same thing that happens when I run `zk new` from that directory.

For what it's worth, `zk edit ./ --interactive` (or `zk edit journal --interactive`) resolves the group correctly, so passing an explicit path works around the problem — only the no-path invocation is affected. And `zk new` run from inside the group directory works fine, so the group config and template themselves are fine.

### How to reproduce?

1. Create a notebook with a group that has its own template, and no template file for the notebook root:

   ```toml
   [group.journal]
   paths = ["journal"]

   [group.journal.note]
   template = "journal.md"
   ```

   with `.zk/templates/journal.md` existing and no `default.md`.
2. `cd journal/`
3. Run `zk edit --interactive`, type a query, and press Ctrl-E ("create a note with the query as title").
4. The note creation fails with `zk: error: new note: cannot locate template at default.md`, followed by `Found 0 notes.` — the root config's template, not `journal.md`. The same happens from the notebook root, since the binding always targets the root.

### zk configuration

```toml
[group.journal]
paths = ["journal"]

[group.journal.note]
template = "journal.md"
```

(`.zk/templates/journal.md` exists; no root template file.)

### Environment

```bash
zk 0.15.4
system: Linux (installed via nix)
```

skbolton:
The "create a note" fzf binding of `zk edit --interactive` always created notes with the notebook root as target, so the root config was used even when invoked from a group directory.

Default to the working directory, like `zk new` does, and fall back to the notebook root when it lies outside the notebook. Explicit path arguments keep the previous behavior.

Fixes #767


docs/notes/note-creation.md
```diff
@@ -22,7 +22,9 @@ existing note or creating a new one in a single action.
 
 This option is available when running `zk edit --interactive`, which spawns
 [`fzf`](../config/tool-fzf.md) to filter selected notes. From `fzf`, press `Ctrl-E` to
-create a new note using the current search query as title.
+create a new note using the current search query as title. The note is created
+in the directory `zk edit` was run from, or in the directory passed as
```
WhyNotHugo: This should not be true if the current directory is outside the notebook.

internal/cli/cmd/edit_test.go
```diff
@@ -0,0 +1,78 @@
...
+	for _, dir := range []string{"journal", ".zk/templates"} {
+		err := os.MkdirAll(filepath.Join(notebookDir, dir), 0o700)
+		assert.Equal(t, err, nil)
+	}
+
+	write := func(path, content string) {
+		err := os.WriteFile(filepath.Join(notebookDir, path), []byte(content), 0o600)
+		assert.Equal(t, err, nil)
+	}
+	write(".zk/templates/journal.md", "Journal: {{title}}\n")
```
WhyNotHugo: This function is used exactly once and can be inlined.

internal/cli/cmd/edit_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+	newNotebook := func(workingDir string) *core.Notebook {
```
WhyNotHugo: Same here, the indirection is unnecessary.

internal/cli/cmd/edit_test.go
```diff
@@ -0,0 +1,78 @@
...
+		note, err := createNote(t, notebookDir+"/journal")
+		assert.Equal(t, err, nil)
+		assert.Equal(t, note.Path, "journal/test.md")
+		assert.Equal(t, note.RawContent, "Journal: Test\n")
+	})
+
+	t.Run("a note created from the notebook root uses the root config", func(t *testing.T) {
+		_, err := createNote(t, notebookDir)
+		assert.Err(t, err, "cannot locate template at default.md")
+	})
```
WhyNotHugo: Should also test when working directory is outside notebook.

WhyNotHugo: This block would probably be much simpler as https://go.dev/wiki/TableDrivenTests

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ Format: `<description> (by <contributor>, <pr number>)`
 
 - Resolve links with a leading slash against the notebook root, falling back to
   the filesystem for absolute paths (by @zmre, 745)
+- Use the working directory as the target of the `edit --interactive` new note
+  fzf binding, so the config group of the current directory is honored (by
+  @skbolton, 768)
```
WhyNotHugo:
```suggestion
- Use the working directory as the target of the `edit --interactive` new note
  fzf binding (by @skbolton, 768)
```

WhyNotHugo (changes requested):

internal/cli/cmd/edit_test.go
skbolton: Fair. I initially had more tests in place but pulled them back. I'll make this adjustment

internal/cli/cmd/edit_test.go
skbolton: Sounds good.

internal/cli/cmd/edit_test.go
skbolton: Okay, I'll look into that

docs/notes/note-creation.md
skbolton: I think I'll drop this doc update all together since I think what is there is good enough.

skbolton:
Pushed another commit with the suggested changes. I didn't want to rebase since github can get wonky with reviews on rebases so when this gets approved I can do another push that squashes if its easier on the maintainers, or they can squash for me to generate the final commit. Whatever is easier.

internal/cli/cmd/edit_test.go
skbolton: Fixed

internal/cli/cmd/edit_test.go
skbolton: Fixed

internal/cli/cmd/edit_test.go
skbolton: Implemented

CHANGELOG.md
skbolton: Fixed

WhyNotHugo (approved):
LGTM

